Issue #19, opened 2023-05-08
Fully support person media_type


IRHM:
Fully support people type from tmdb.

- [x] Fix image not being shown on poster.
- [x] Have a differentiator so we know when a poster is for a person, not actual media.
- [x] Person details page, with their known for movies?

PR #33, opened 2023-05-16
Improve and Separate Person Posters

IRHM:
- Created `PersonPoster` component and removed its logic from `Poster`
- Improved design of person posters
- Make addClassToParent a helper method
- Posters: Break word if name too long


IRHM:
Finalizes #19

PR #111, opened 2023-08-20
add year to Poster

### LINKED ISSUE #108: Poster: Add content year
The year of movie/show should be shown on the poster on hover (alongside title?).

kurtmgray:
Added year under title, taking style cues from ./content/Title.svelte. Checked both release_year and first_air_date properties of movie to render a year. #108


src/lib/Poster.svelte
```diff
@@ -38,6 +40,8 @@
   const title = media.title || media.name;
   const poster = `https://image.tmdb.org/t/p/w500${media.poster_path}`;
   const link = media.id ? `/${media.media_type}/${media.id}` : undefined;
+  const dateStr = media.release_date || media.first_air_date;
+  const year = dateStr ? new Date(dateStr).getFullYear() : undefined;
```
IRHM:
```suggestion
  const year = new Date(media.release_date || media.first_air_date)?.getFullYear();
```

Not an important change, but I think it can help simplify (of course if you make this change, need to remove the `dateStr` variable too).

I removed the use of the ternary operator (for year), It is probably not worth setting it to undefined, if the method doesn't work, it will be already.

src/lib/Poster.svelte
```diff
@@ -121,6 +125,7 @@
           {title}
         {/if}
       </h2>
+      <time>{year}</time>
```
IRHM: This might look better if its put next to the title of the movie (like how it is in the movie or tv page title):

image

IRHM:
```suggestion
     {#if year}
        <time>{year}</time>
      {/if}
```

We should probably put an {#if} around this, so if it is `undefined`, it won't show.

src/lib/Poster.svelte
```diff
@@ -20,7 +20,9 @@
     name?: string;
     overview?: string;
     id: number; // tmdb id
-    media_type: MediaType;
+    media_type: MediaType; 
+    release_date?: string;
```
IRHM, line 24: This is good, though it doesn't work on the main watched list page (`src/routes/(app)/+page.svelte`).

The Poster media prop is passed manually in there, so something like this needs to be added there (in the media object):
```
release_date: w.content.release_date
```

I think you may also need to update the Content type in types.ts to include release_date.

IRHM (changes requested):
Thanks for this PR.

I made a few suggestions that could improve the implementation.

If you need any help or are stuck on anything, feel free to reply!

src/lib/Poster.svelte
kurtmgray: Thanks for all the feedback. I'll implement the changes asap.

src/lib/Poster.svelte
kurtmgray: My understanding is that passing in an undefined value, as we would here `new Date(undefined || undefined)` would result in an invalid date. TS also doesn't allow `undefined` as an argument to new Date(). I'd propose `const year = (media. release_date || media. first_air_date) && new Date ((media. release_date || media. first_air_date) !) .getFullYear) ;` if you want to eliminate the dateStr variable, though it seems a bit duplicative. Thoughts?

src/lib/Poster.svelte
kurtmgray: Great, done!

src/lib/Poster.svelte
kurtmgray: Done!

src/lib/Poster.svelte
IRHM: @kurtmgray Ah I see, we should probably just keep it how it was then in that case.

It was good at first, I thought this might have been a small improvement. Neglected to think what typescript would have to say about it.

ps my bad 🫥

IRHM:
Thank you for the updates! They are looking very good.

Another thing to add, it looks like a few images and your log file got checked into the repo by accident. You can move your data folder to /server/data (included path in .gitignore) or you can feel free to add another entry into .gitignore to where you data is now (may need to remove the files, do a commit, then update .gitignore and re-add the files).

kurtmgray:
Thanks for pointing that out. All changes made.

src/lib/Poster.svelte
kurtmgray: Great, not a problem. I went with `const year = dateStr && new Date(dateStr).getFullYear();` which will give `year` the `string | undefined` type we want, dependent upon `dateStr`.

IRHM (approved):
Looks great, thank you for working on this! :rocket:

PR #119, opened 2023-08-24
TMDB Request Caching

### LINKED ISSUE #59: Cache TMDB Requests
The server should cache every TMDB request for a certain amount of time (24h sounds reasonable?).

There is probably a middleware/package available that can handle this out the box for us, check before implementing ourselves.

IRHM:
Implement caching of external tmdb api requests (24h).

Closes #59

### DIFF
--- a/server/go.mod
+++ b/server/go.mod
@@ -3,6 +3,7 @@ module github.com/sbondCo/Watcharr
 go 1.21
 
 require (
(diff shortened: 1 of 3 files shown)


IRHM:
im scared how easy this seems to have been to add with the gin cache plugin, hopefully it isn't secretly broken.

Log seems to show no tmdb requests going out whilst our cache is available.

PR #126, opened 2023-08-28
Make Nav Sticky

### LINKED ISSUE #121: Sticky Nav
The nav bar should be sticky.

If easy, the nav bar should probably hide when scrolling down, then appear again when scrolling up.

We don't want it to pop in and out of existence though, so when you scroll up, it should scroll into view.

IRHM:
Hides on scroll down, comes into view on scroll up.

Closes #121

### DIFF
--- a/src/routes/(app)/+layout.svelte
+++ b/src/routes/(app)/+layout.svelte
@@ -212,9 +212,14 @@
     display: flex;
     align-items: center;
     justify-content: space-between;
-    margin: 10px 20px 28px 20px;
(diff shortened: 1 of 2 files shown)


IRHM:
Leaving this open for a bit until I decide if i like it better sticky or not :disappointed:

IRHM:
> Leaving this open for a bit until I decide if i like it better sticky or not 😞

@IRHM Two side by side gifs to compare the before and after would definitely help you decide!

PR #131, opened 2023-09-11
Do not fail on `.env` file missing

stavros-k:
In setups where you set the env vars directly, eg Docker or Kubernetes, there is no need to require a .env file

There already the `ensureEnv` below that makes sure required variables are there,

### DIFF
--- a/server/watcharr.go
+++ b/server/watcharr.go
@@ -35,7 +35,10 @@ var (
 func main() {
 	err := godotenv.Load()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


IRHM (approved):
Thanks for the contribution! Good catch on this, you're right that we already verify if the required environment variables are present.

My pc doesn't work right now 😵, but I'll see what I can do tomorrow to get a patch release out for this to support your use case.

stavros-k:
Take your time! Looks like those go courses paid up lol

PR #132, opened 2023-09-17
Importing text watch list

### LINKED ISSUE #65: Import from text file
Probably want a delay between each new item so we dont 'spam' tmdb (not for cached content of course)

A UI for the import would be nice, we can check off each item as we import them. We can also then show errors for the content we couldn't find and allow for the user to resolve the issue, either by correcting the name or manually inputting an id from tmdb.

Import button should be added to empty watch list menu (for current user only)

- [x] Import files
  - [x] Click to browse
  - [x] Drag and drop
- [x] Display imported file data
  - [x] Text file (supports extracting year from entry in text file if between brackets)
  - [x] #130
- [x] Allow editing displayed import data
  - [x] Edit name
  - [x] Edit Type
  - [x] Add new row
  - [x] Remove row
- [x] Importing the actual data one by one
  - [x] Support showing multiple results for import and allowing user to decide
    - [x] Create modal view
    - [x] Click poster in modal and try import again
    - [x] Allow skipping selection (closing modal should skip too.. update with error icon on table)
    - [ ] Allow providing tmdb id manually in modal
  - [x] Success, error, ? icons state updates for table
- [x] New activity type of Import (instead of added at, we can show imported at)
- [x] Add import button to empty watched lists
- [x] Make sure works in both light and dark themes

IRHM:
A way to import content from text files into your watched list.


IRHM:
This pr is getting way too huge, didn't anticipate the amount of work involved. It is marked as a beta feature on the import page for now.

Going to merge now, and when issues are eventually found tomorrow, they can be fixed in separate PRs.

PR #214, opened 2023-11-19
Tmdb import

MathieuMoalic:
This is WIP, I have issues with adding Activities on the server. 

The frontend code is quite straight forward and was easy to implement. 
The backend was something else. I found the functions to be really large and do many things, it was hard to wrap my head around it. Instead of changing the existing import route, I made a parallel one "/tmdb_import", where I could freely tinker the code without fears of breaking what you had already made. The parsing and everything loads perfectly but I can't figure out at all why adding activities doesn't do anything. Example from `tmdb_import.go`:

```go
var activity Activity
activityJson, err := json.Marshal(map[string]interface{}{"status": addRequest.Status, "rating": addRequest.Rating})
if err != nil {
slog.Error("Failed to marshal json for data in ADD_WATCHED activity request, adding without data", "error", err.Error())
activity, _ = addActivity(db, userId, ActivityAddRequest{WatchedID: uint(addRequest.ContentID), Type: at})
} else {
activity, _ = addActivity(db, userId, ActivityAddRequest{WatchedID: uint(addRequest.ContentID), Type: at, Data: string(activityJson)})
}
``` 
There are no errors, but it just doesn't do anything. I also have a lot of issues with the logger, only `slog.Info` seem to show the messages. Forgive my ignorance, I never trained as a software dev ^^'

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,6 @@ vite.config.ts.timestamp-*
 
 # Docker dev volume
 container_data/
+flake.lock
+flake.nix
+pnpm-lock.yaml
--- a/package.json
+++ b/package.json
@@ -32,6 +32,9 @@
   },
   "type": "module",
   "dependencies": {
-    "axios": "^1.6.0"
+    "@types/papaparse": "^5.3.11",
+    "axios": "^1.6.0",
+    "papaparse": "^5.4.1",
+    "sass": "^1.69.5"
   }
 }
--- a/server/content.go
+++ b/server/content.go
@@ -14,7 +14,7 @@ const (
 )
 
 // For storing cached content, so we can serve the basic local data for watched list to work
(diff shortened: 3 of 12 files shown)


IRHM:
Hey, thanks for making the PR. I will have a look when i get back home later today.

Sorry to hear you had issues, not too sure why the logger was having issues. I suspect the activity may have been adding with the wrong id (content id instead of watched item id?), and so was not showing for the watched item.

Thanks again I'll check it out soon.

IRHM:
@MathieuMoalic Thank you for sharing the progress, I was able to rework into a client side only implementation.

Thought it made more sense to make a new PR (#216), co-authored the commits with you since you did the starting work.

Thank you again! I'll look at testing on dev and releasing it soon.

Issue #130, opened 2023-09-08
Importing exported list


IRHM:
After #66 is done, then we can add importing based on its format(s).

MathieuMoalic:
Hi, are there plans to add the option to import from tmdb? The formatting looks like this and is not compatible with the current import feature:
```
TMDb ID,IMDb ID,Type,Name,Release Date,Season Number,Episode Number,Rating,Your Rating,Date Rated
429,tt0060196,movie,"Il buono, il brutto, il cattivo",1966-12-23T00:00:00Z,,,8.473,9.0,2014-11-08T00:00:00Z
155,tt0468569,movie,The Dark Knight,2008-07-16T00:00:00Z,,,8.512,10.0,2014-11-08T00:00:00Z
27205,tt1375666,movie,Inception,2010-07-15T00:00:00Z,,,8.365,9.0,2014-11-08T00:00:00Z
```

IRHM:
Hi @MathieuMoalic, it wasn't but I can definitely have a look later tonight to how big of a task it would be to add support for that format (looks like csv).

Currently it only supports importing by name of movie/show, but having the tmdb id should make it easy.

MathieuMoalic:
Good to hear, I can try to make it myself if you want help.

IRHM:
> Good to hear, I can try to make it myself if you want help.

That would be very helpful, thank you! If you need anything you can let me know.

IRHM:
Hi @MathieuMoalic, did you have any time to look into this?

With the last version out, I have more time to look into other tasks, I can look into this one for you if you haven't yet or if you need anything.

MathieuMoalic:
Hi, I have done some work on it, of course it's more work than expected. I will open an incomplete PR soon to share the progress.

IRHM:
> Hi, I have done some work on it, of course it's more work than expected. I will open an incomplete PR soon to share the progress.

Hah yeah, there must be a law where the real estimate of work is * 10 the estimate. Thank you for taking the time to look into this, whatever has been done will surely be very helpful!

stephaje:
Is there any progress on this issue? I would be interested in taking it on if there is nothing in progress already.

PR #216, opened 2023-11-19
TMDb Importing

### LINKED ISSUE #215: Support TMDB/IMDB importing

IRHM:
Add support for importing from TMDb watched list csv export.

Related #214

Closes #215

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,11 +8,13 @@
       "name": "watcharr",
       "version": "1.23.0",
       "dependencies": {
-        "axios": "^1.6.1"
+        "axios": "^1.6.1",
+        "papaparse": "^5.4.1"
       },
       "devDependencies": {
         "@sveltejs/adapter-node": "^1.3.1",
         "@sveltejs/kit": "^1.27.5",
+        "@types/papaparse": "^5.3.11",
         "@typescript-eslint/eslint-plugin": "^6.11.0",
         "@typescript-eslint/parser": "^6.11.0",
         "eslint": "^8.53.0",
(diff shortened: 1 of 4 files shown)


MathieuMoalic:
Nice one, you found a much cleaner solution than what I could do :) . I do have a couple questions though, when I try to import, it doesn't keep the ratings, is it a bug or intended?

IRHM:
> Nice one, you found a much cleaner solution than what I could do :) . I do have a couple questions though, when I try to import, it doesn't keep the ratings, is it a bug or intended?

Oh I didnt implement saving ratings.

When I was testing creating a list on tmdb, for some reason when I gave something a rating it would be removed from my list.

I see now that your example data (https://github.com/sbondCo/Watcharr/issues/130#issuecomment-1795856431) has a YourRating.

I think I saw you could add notes to movies too, but that seems to be on custom lists you cant export on tmdb. So I guess we cant do that to save `thoughts` too.

I'll have a look at adding the ratings, should be an ezpz fix (hopefully).

MathieuMoalic:
Also, the data includes the date of the rating, I think it would be cool not to lose that as well.

IRHM:
> Also, the data includes the date of the rating, I think it would be cool not to lose that as well.

Hey, just released [v1.25.0](https://github.com/sbondCo/Watcharr/releases/tag/v1.25.0) with this functionality.

Changing the date of an activity is a planned feature so I was able to get some of that work done in the backend for saving the original rating date.

It'll look something like this after importing (bottom activity says "Original Rating on TMDb" - my screenshot tool buggy :disappointed:):

image


Let me know if anything else comes up!

PR #223, opened 2023-11-20
Import rating and rating date from TMDb, give activity customDate column

IRHM:
- Import ratings from TMDb export
- Give activity a `customDate` column. Activity dates will be editable in the future, so decided to add this now to solve this problem.

Related #216

### DIFF
--- a/server/activity.go
+++ b/server/activity.go
@@ -3,6 +3,7 @@ package main
 import (
 	"errors"
 	"log/slog"
+	"time"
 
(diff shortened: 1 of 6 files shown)


MathieuMoalic:
Works perfectly! ( exept for individual episodes rating but I personally don't use that ). It also gives some network_changed errors randomly when processing a long list of imports but it looks like the imports all went through anyway.
Thanks for the work :+1:

PR #255, opened 2023-12-08
Hiding rating when hide spoiler is checked in profile

### LINKED ISSUE #217: SeasonsList: Hide Spoiler setting should also hide the episode rating
SeasonsList: Hide Spoiler profile setting should also hide the episode rating.

The rating of an episode could spoil if its a good or bad episode. Watching a series and seeing the next episode has a 0.0002 stars out of 10 kinda spoils that it is trash.

We could add a separate button to only unhide the rating for people who dont want to waste time on badly rated episodes, without revealing thumbnail.

ishaanparlikar:
Fixes #217

### DIFF
--- a/src/lib/SeasonsListEpisode.svelte
+++ b/src/lib/SeasonsListEpisode.svelte
@@ -138,6 +138,9 @@
       img {
         filter: blur(6px);
       }
+      .rating {
+        filter:blur(4px)
+      }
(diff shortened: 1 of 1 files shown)


IRHM:
@ishaanparlikar Thank you for doing this task, I appreciate it!

PR #259, opened 2023-12-09
Feature/pwa

### LINKED ISSUE #225: PWA support
Look into making Watcharr a PWA.

ishaanparlikar:
Added PWA feature.

Closes #225

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,8 +30,7 @@
     "svelte-preprocess": "^5.1.1",
     "tslib": "^2.6.2",
     "typescript": "^5.3.2",
-    "vite": "^4.5.1",
-    "workbox-window": "^7.0.0"
+    "vite": "^4.5.1"
   },
   "type": "module",
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "@types/papaparse": "^5.3.14",
     "@typescript-eslint/eslint-plugin": "^6.13.0",
     "@typescript-eslint/parser": "^6.13.1",
+    "@vite-pwa/sveltekit": "^0.2.10",
     "eslint": "^8.55.0",
     "eslint-config-prettier": "^9.1.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -29,7 +30,8 @@
     "svelte-preprocess": "^5.1.1",
     "tslib": "^2.6.2",
     "typescript": "^5.3.2",
-    "vite": "^4.5.1"
+    "vite": "^4.5.1",
+    "workbox-window": "^7.0.0"
```
IRHM: @ishaanparlikar Looks like we don't use this dependency (workbox-window)? Are we able to remove it?

vite.config.ts
```diff
@@ -1,11 +1,39 @@
 import { sveltekit } from "@sveltejs/kit/vite";
 import { defineConfig } from "vite";
 import { readFileSync } from "fs";
-
+import {SvelteKitPWA} from '@vite-pwa/sveltekit'
+/** @type {import('vite').UserConfig} */
```
IRHM: Can we remove this `@type` line?

IRHM (changes requested):
Found a dependency I don't think is being used?

~Were you able to get the app to install in your browser? I think there should be an Install button in Firefox/chromium, but I don't see one anywhere. Not sure if I'm doing something wrong here.~

package.json
ishaanparlikar: @IRHM  I was getting error before installing this dependancy. But I checked just now, I was able to build the project without the issue. I have removed it. Please have a look.
Thanks

vite.config.ts
ishaanparlikar: @IRHM  Removed.

IRHM (approved):

IRHM:
@ishaanparlikar Looks great, thanks again for getting this task done! I'll test and release this with the next release later tonight or tomorrow.

Issue #276, opened 2023-12-20
Allow filtering and sorting other watched lists


IRHM:
We should allow filtering and sorting of other peoples watched lists.

iamericfletcher:
If this is still unassigned, I'd like to give this a shot! I've got the project running locally on my Linux machine running Pop!_OS (Using Source instead of Docker for now). This will be my first time using `Go` and `Svelte` but I think I'm up for the task :)

I have all next week off from work so plan on studying the code base and coming up with a description of my approach to add here for review and discussion. 

Is this OK?

My first question is:

1. Do you have any recommendations for improving the development environment such that say changes to the front end trigger a reload on the server without having to rebuild? 

For example, say I change the below in `+page.svelte`:

`
    <h2>
      {#if login}
        Get Back In!
      {:else}
        Lucky You Found Us!
      {/if}
    </h2>
`

to:

`
    <h2>
      {#if login}
        Get Back In!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
      {:else}
        Lucky You Found Us!
      {/if}
    </h2>
`

I'd like to then reload the server and see those changes reflected without having to rebuild. 

Is this possible? Also, apologies for what may seem like a beginner question! I hope it doesn't dissuade you from assigning the issue to me!

IRHM:
Hi @iamericfletcher, thanks for taking an interest, that would be great!

Just to let you know, the filtering/sorting of your own watched list is all done client side (svelte/js), having a look at that existing code will surely help, I'm sure most of it can be reused.

Thanks again, if you need anything, feel free to let me know, I'm happy to help!

iamericfletcher:
@IRHM - That's what I was thinking - that most of the existing filtering/sorting code can be reused :) A great first issue!

IRHM:
@iamericfletcher, dont worry about it.

I dont have anything setup to rebuild the backend automatically, but for rebuilding the svelte code, running: `npm run dev` should setup a dev server with live reloading.

You can check out the package.json file for the other "scripts" that can be helpful during development (npm run <script name>).

iamericfletcher:
Awesome - thanks so much and what an awesome project this is! Looking forward to coming up with a plan and perhaps some code for your review / feedback in the near future. 

Sincerely,
Eric

IRHM:
Completed in #281, thanks!

PR #281, opened 2023-12-24
Enhancement: Allow filtering and sorting other watched lists #276

iamericfletcher:
### Changes made

Added a new store variable 'watchedListErrorOccurred' set to true if `getPublicWatchedList()` promise errors.

The aim is that the filter and sort functionality only show if the watched list loads in successfully.

### DIFF
--- a/src/routes/(app)/+layout.svelte
+++ b/src/routes/(app)/+layout.svelte
@@ -1,7 +1,6 @@
 <script lang="ts">
   import { goto } from "$app/navigation";
   import { page } from "$app/stores";
-  import { watchedListErrorOccurred } from "@/store";
   import Icon from "@/lib/Icon.svelte";
   import PageError from "@/lib/PageError.svelte";
   import Spinner from "@/lib/Spinner.svelte";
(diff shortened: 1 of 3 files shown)


IRHM (changes requested):
Hi @iamericfletcher, thank you for getting this working!

There seems to be an issue where if `watchedListErrorOccurred` is set to true and I navigate back to another profile, it never gets set back to false on successful load of a watched list.

I'm also not sure if the added complexity of `watchedListErrorOccurred` is worth it? It might be a worse UX having buttons fully disappear in certain conditions.

Having the buttons disable would be a better option, but since the sort/filter menus work based off of global `activeSort` and `activeFilters` stores, it may be better leaving them enabled in all cases.

iamericfletcher:
Hi @IRHM - I totally agree. Upon further consideration, `watchedListErrorOccurred` does indeed add unnecessary complexity. I've updated the code accordingly. I've made the changes and pushed the new commit just now for your review and feedback.

IRHM:
Thank you very much @iamericfletcher, this works great!

PR #290, opened 2023-12-27
Implement Change Password Feature (Issue #279)

### LINKED ISSUE #279: Change password
New button in profile settings to allow the user to change their password. Probably should open a new modal that allows the user to input their `current password` and a `new password`.

iamericfletcher:
### Changes made

Implemented new change password feature. 

- Created new `PwChangeModal` and added it to `src/routes/profile/modals/PwChangeModal.svelte`
- Created `changeUserPassword` in `src/lib/util/api.ts`
- Created `POST("/change_password")` with `addUserRoutes` in `server/routes.go` - so `/user/change_password`
- Created `userChangePassword` in `server/user.go` with `UserPasswordUpdateRequest` struct in `server/auth.go`

Thoughts:

- Could probable create a `FormList` similar to `SettingsList` (`src/lib/settings/SettingsList.svelte`)

```
  .form-input-container {
    display: flex;
    flex-flow: column;
    gap: 20px;
    margin: 0 15px;
  }

  .form-input-container > button {
    margin-left: auto;
    width: max-content;
  }
```


server/user.go
```diff
@@ -164,3 +164,46 @@ func uploadUserAvatar(c *gin.Context, db *gorm.DB, userId uint) (Image, error) {
 	}
 	return img, nil
 }
+
+func userChangePassword(db *gorm.DB, pwds UserPasswordUpdateRequest, userId uint) error {
```
IRHM: This may fit better in `auth.go` (This function is more related to auth than what else is in `user.go`). The `user.go` file is more for editing settings or finding other users.

server/user.go
```diff
@@ -164,3 +164,46 @@ func uploadUserAvatar(c *gin.Context, db *gorm.DB, userId uint) (Image, error) {
...
+	slog.Debug("userChangePassword request running", "user_id", userId)
+	user := new(User)
+	res := db.Where("id = ?", userId).Take(&user)
+	if res.Error != nil {
+		slog.Error("userChangePassword failed - failed to retrieve user from database", "user_id", userId, "error", res.Error)
+		return errors.New("failed to retrieve user")
+	} else {
```
IRHM: The above `if` returns, so it acts like a guard. We could help with readability by removing the `else`.

server/user.go
```diff
@@ -164,3 +164,46 @@ func uploadUserAvatar(c *gin.Context, db *gorm.DB, userId uint) (Image, error) {
...
+			memory:      64 * 1024,
+			iterations:  3,
+			parallelism: 2,
+			saltLength:  16,
+			keyLength:   32,
+		})
+		if err != nil {
+			slog.Error("userChangePassword failed - failed to hash new password", "user_id", userId, "error", err)
+			return errors.New("failed to hash new password")
+		} else {
```
IRHM: Same as above. We could help with readability by removing the `else`.

server/user.go
```diff
@@ -164,3 +164,46 @@ func uploadUserAvatar(c *gin.Context, db *gorm.DB, userId uint) (Image, error) {
...
+			slog.Error("userChangePassword failed - failed to compare passwords", "user_id", userId, "error", err)
+			return errors.New("failed to compare passwords")
+		}
+		if !match {
+			slog.Error("userChangePassword failed - current password hash doesn't match password hash in database", "user_id", userId, "error", err)
+			return errors.New("current password provided doesn't match password in database")
+		}
+		slog.Debug("userChangePassword hash for current password matches hash in the database", "user_id", userId)
+		slog.Debug("userChangePassword hashing new password", "user_id", userId)
+		hash, err := hashPassword(pwds.NewPassword, &ArgonParams{
```
IRHM: This is up to you if you want to. Since the `ArgonParams` must be the same here as in the `register` function, we could move them both out to a `func` (in `auth.go`):

```
func GetPassArgonParams() *ArgonParams {
	return &ArgonParams{
		memory:      64 * 1024,
		iterations:  3,
		parallelism: 2,
		saltLength:  16,
		keyLength:   32,
	}
}
```

I'm leaning towards a func (instead of a var) so we can rest easy knowing the parameters can't be changed (for everything).

src/lib/util/api.ts
```diff
@@ -170,6 +170,23 @@ export function updateUserSetting<K extends keyof UserSettings>(
...
+    .then((r) => {
+      if (r.status === 200) {
+        notify({ id: nid, type: "success", text: "Password Changed" });
+        if (typeof done !== "undefined") done();
+      }
+    })
+    .catch((err) => {
+      const errMsg = err.response.data.error ? err.response.data.error : "Couldn't Change Password";
+      console.error("Change Password Form - Failed to change password on the server", err);
+      notify({ id: nid, type: "error", text: errMsg });
```
IRHM: Instead of using `notify` here, we could pass the error to the `done` callback. In the modal, we could use that to set `error` to show an error box at the top.

I think this should make it easier for users, so all errors show up in the same place.

src/routes/(app)/profile/modals/PwChangeModal.svelte
```diff
@@ -0,0 +1,159 @@
...
+  function handleSubmit(ev: SubmitEvent) {
+    checkForm();
+    if (!error) {
+      console.log(
+        "Change Password Form - Status: Form inputs on the frontend are valid - sending request to change password to the server"
+      );
+      const fd = new FormData(ev.target! as HTMLFormElement);
+      changepswd.currentPassword = fd.get("current-password") as string;
+      changepswd.newPassword = fd.get("new-password") as string;
+      changeUserPassword(changepswd.currentPassword, changepswd.newPassword, () => {
```
IRHM: Related to the other comment, it could then be used like so:

```
changeUserPassword(changepswd.currentPassword, changepswd.newPassword, (err) => {
  if (err) {
    error = err;
    return;
  }
  console.log("Change Password Form - Status: Password changed successfully");
  onClose();
});
```

IRHM (changes requested):
Hi @iamericfletcher, this functions very well, thank you for working on it!

I added a few suggestions that I think could improve the changes. Let me know if you have any thoughts/questions about them.

server/user.go
iamericfletcher: I agree! I was on the fence about this myself and am glad you mentioned this :)

server/user.go
iamericfletcher: Totally agree in `ArgonParams` being a `func` instead of `var`!

IRHM (dismissed):
Works great! Thank you again!

I added a few small changes I thought weren't worth hassling you over. Mainly:

- `SELECT` only the password from database
- Use `?` where we set `errMsg` to avoid erroring if `err`, `response` or `data` are undefined.

iamericfletcher:
> Works great! Thank you again!
> 
> I added a few small changes I thought weren't worth hassling you over. Mainly:
> 
> * `SELECT` only the password from database
> * Use `?` where we set `errMsg` to avoid erroring if `err`, `response` or `data` are undefined.

You're very welcome! Learning a lot from you and this project, so thank you :)

PR #302, opened 2024-01-07
Watchtime Profile Stat

### LINKED ISSUE #256: Profile Stat: Calculate users total watchtime of movies, tv, combined
3 new stats for the profile page.

Total time spent watching:
- Movies
- Shows
- Both (sum of previous two)

Calcilating for movies should be easy, we already store runtimes.

Shows will be more complicated since tmdb started transitioning from array of possible runtimes to having a runtime for each episode of a show (i think i read this looking into a different issue). Get any runtime available and * by number of episodes. This change may mean we have missing runtimes in db.

Both will be the sum of movies and shows.

IRHM:
### Changes made
Movie and Show watchtime stats in profile page.

image

Closes #256

### DIFF
--- a/server/profile.go
+++ b/server/profile.go
@@ -13,6 +13,7 @@ type Profile struct {
 	ShowsWatched         int32     `json:"showsWatched"`
 	MoviesWatched        int32     `json:"moviesWatched"`
 	MoviesWatchedRuntime uint32    `json:"moviesWatchedRuntime"`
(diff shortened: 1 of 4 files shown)


IRHM:
Show total watchtimes will for most shows going forward probably be very inaccurate, which will mess with these stat totals. Need to come back and fix this if I can.

PR #366, opened 2024-02-18
Issue #61

### LINKED ISSUE #61: Manage activity
Allow user to manage their content activity manually.

Editing isn't required, but deleting and adding should be added (timestamp for activity should be modifiable)

stephaje:
### Changes made

- Allow clicking on activity entries to bring up an editor
- Allow for timestamp updating and deletion of activity entries
- Allow for different sized modals
- Misc: add build artifacts to gitignore
- Misc: change activity text to be in proper title case
- Misc: fix small slog missing key warning

### DIFF
--- a/server/activity.go
+++ b/server/activity.go
@@ -81,27 +81,27 @@ func addActivity(db *gorm.DB, userId uint, ar ActivityAddRequest) (Activity, err
 	return activity, nil
 }
 
-func updateActivity(db *gorm.DB, id uint, activityUpdateRequest ActivityUpdateRequest) error {
+func updateActivity(db *gorm.DB, userId uint, id uint, activityUpdateRequest ActivityUpdateRequest) error {
 	if id == 0 {
 		return errors.New("id must be set to update an activity")
 	}
 	if activityUpdateRequest.CustomDate.IsZero() {
 		return errors.New("customDate must be set to update an activity")
 	}
-	res := db.Model(&Activity{}).Where("id = ?", id).Update("custom_date", activityUpdateRequest.CustomDate)
+	res := db.Model(&Activity{}).Where("user_id = ? AND id = ?", userId, id).Update("custom_date", activityUpdateRequest.CustomDate)
 	if res.Error != nil {
-		slog.Error("Error deleting activity in database", "error", res.Error.Error())
-		return errors.New("failed deleting activity in database")
+		slog.Error("Error updating activity in database", "error", res.Error.Error())
+		return errors.New("failed updating activity in database")
 	}
-	slog.Debug("Deleting activity", "deleted_activity", id)
+	slog.Debug("Updating activity", "updated_activity", id)
 	return nil
 }
 
-func deleteActivity(db *gorm.DB, id uint) error {
+func deleteActivity(db *gorm.DB, userId uint, id uint) error {
 	if id == 0 {
 		return errors.New("an id must be provided to delete an activity")
 	}
-	res := db.Delete(&Activity{}, id)
+	res := db.Where("user_id = ?", userId).Delete(&Activity{}, id)
 	if res.Error != nil {
 		slog.Error("Error deleting activity in database", "error", res.Error.Error())
 		return errors.New("failed deleting activity in database")
--- a/server/activity.go
+++ b/server/activity.go
@@ -53,6 +53,10 @@ type ActivityAddRequest struct {
 	CustomDate *time.Time   `json:"customDate,omitempty"`
 }
 
+type ActivityUpdateRequest struct {
+	CustomDate time.Time `json:"customDate" binding:"required"`
+}
+
 func getActivity(db *gorm.DB, userId uint, watchedId uint) ([]Activity, error) {
 	activity := new([]Activity)
 	res := db.Model(&Activity{}).Where("user_id = ? AND watched_id = ?", userId, watchedId).Find(&activity)
@@ -75,4 +79,32 @@ func addActivity(db *gorm.DB, userId uint, ar ActivityAddRequest) (Activity, err
 	}
 	slog.Debug("Adding activity", "added_activity", activity)
 	return activity, nil
+}
+
+func updateActivity(db *gorm.DB, id uint, activityUpdateRequest ActivityUpdateRequest) error {
+	if id == 0 {
+		return errors.New("id must be set to update an activity")
+	}
+	if activityUpdateRequest.CustomDate.IsZero() {
+		return errors.New("customDate must be set to update an activity")
+	}
+	res := db.Model(&Activity{}).Where("id = ?", id).Update("custom_date", activityUpdateRequest.CustomDate)
+	if res.Error != nil {
+		slog.Error("Error deleting activity in database", "error", res.Error.Error())
+		return errors.New("failed deleting activity in database")
+	}
+	slog.Debug("Deleting activity", "deleted_activity", id)
+	return nil
+}
+
+func deleteActivity(db *gorm.DB, id uint) error {
+	if id == 0 {
+		return errors.New("an id must be provided to delete an activity")
+	}
+	res := db.Delete(&Activity{}, id)
+	if res.Error != nil {
+		slog.Error("Error deleting activity in database", "error", res.Error.Error())
+		return errors.New("failed deleting activity in database")
+	}
+	return nil
 }
--- a/server/activity.go
+++ b/server/activity.go
@@ -93,6 +93,10 @@ func updateActivity(db *gorm.DB, userId uint, id uint, activityUpdateRequest Act
 		slog.Error("Error updating activity in database", "error", res.Error.Error())
 		return errors.New("failed updating activity in database")
 	}
+	if res.RowsAffected < 1 {
+		slog.Error("No activities were updated. This may be because the activity doesn't exist or is not owned by the calling user.")
+		return errors.New("failed updating activity in database")
+	}
 	slog.Debug("Updating activity", "updated_activity", id)
 	return nil
 }
@@ -105,6 +109,10 @@ func deleteActivity(db *gorm.DB, userId uint, id uint) error {
 	if res.Error != nil {
 		slog.Error("Error deleting activity in database", "error", res.Error.Error())
 		return errors.New("failed deleting activity in database")
+	}
+	if res.RowsAffected < 1 {
+		slog.Error("No activities were deleted. This may be because the activity doesn't exist or is not owned by the calling user.")
+		return errors.New("failed deleting activity from database")
 	}
 	return nil
 }
--- a/server/routes.go
+++ b/server/routes.go
@@ -516,6 +516,7 @@ func (b *BaseRouter) addActivityRoutes() {
 			context.Status(http.StatusOK)
 			return
 		}
+		context.AbortWithStatusJSON(http.StatusBadRequest, ErrorResponse{Error: err.Error()})
 	})
 
 	activity.DELETE(":id", func(context *gin.Context) {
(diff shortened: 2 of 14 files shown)


server/activity.go
```diff
@@ -76,3 +80,31 @@ func addActivity(db *gorm.DB, userId uint, ar ActivityAddRequest) (Activity, err
...
+func updateActivity(db *gorm.DB, id uint, activityUpdateRequest ActivityUpdateRequest) error {
+	if id == 0 {
+		return errors.New("id must be set to update an activity")
+	}
+	if activityUpdateRequest.CustomDate.IsZero() {
+		return errors.New("customDate must be set to update an activity")
+	}
+	res := db.Model(&Activity{}).Where("id = ?", id).Update("custom_date", activityUpdateRequest.CustomDate)
+	if res.Error != nil {
+		slog.Error("Error deleting activity in database", "error", res.Error.Error())
```
IRHM: Log and error messages here (two below this one as well) should say updating activity.

server/activity.go
```diff
@@ -76,3 +80,31 @@ func addActivity(db *gorm.DB, userId uint, ar ActivityAddRequest) (Activity, err
 	slog.Debug("Adding activity", "added_activity", activity)
 	return activity, nil
 }
+
+func updateActivity(db *gorm.DB, id uint, activityUpdateRequest ActivityUpdateRequest) error {
```
IRHM: Is there anything in place to ensure an authenticated user can only update their own activity?

server/activity.go
```diff
@@ -76,3 +80,31 @@ func addActivity(db *gorm.DB, userId uint, ar ActivityAddRequest) (Activity, err
...
+	res := db.Model(&Activity{}).Where("id = ?", id).Update("custom_date", activityUpdateRequest.CustomDate)
+	if res.Error != nil {
+		slog.Error("Error deleting activity in database", "error", res.Error.Error())
+		return errors.New("failed deleting activity in database")
+	}
+	slog.Debug("Deleting activity", "deleted_activity", id)
+	return nil
+}
+
+func deleteActivity(db *gorm.DB, id uint) error {
```
IRHM: Same as above; is there anything in place to ensure an authenticated user can only delete their own activity (user_id column check)?

src/lib/Activity.svelte
IRHM, line 1: How come all of the string casing has been changed? The CSS handles the casing with `text-transform: capitalize;`, so it's easier to just keep it lowercase?

IRHM (changes requested):
Hey @stephaje, thanks a lot for working on this, I really appreciate it!

I didn't have enough time to fully look through and run the changes right now, but I read over the code on github and added some comments.

server/activity.go
stephaje: Good point, I'm still getting my bearings. Will update.

src/lib/Activity.svelte
stephaje: I don't agree that's the best way to handle it.

- The title and description (at least in this context) should be in title case. Note that `text-transform: capitalize;` is not title case and also would lead to capitalization changes in stylized titles (e.x. [UFOria](https://www.imdb.com/title/tt0081674/?ref_=kw_li_tt), [ronny/lilly](https://www.imdb.com/title/tt8956332/?ref_=ttep_ep5)) if present.
- Therefore, the user of a Modal should be providing the strings in the exact format they want displayed.
- Converting a sentence from lower case to title case via a function is a little annoying, but [not impossible](https://github.com/gouch/to-title-case/blob/master/to-title-case.js), but also suffers from the issue of not respecting stylized titles.

Given all of the above, it's just more convenient to put the strings in title case manually.

IRHM: @stephaje Thanks for explaining, we don't display any stylized titles/text in activity text, but I see what you mean and it does look better.

I have another note (hope they are not annoying), but I'll add it directly on the code so it's less confusing in a moment.

src/lib/Activity.svelte
```diff
@@ -15,94 +19,101 @@
...
+          return `${seasonAndEpToReadable(data.season, data.episode)} Removed`;
         }
-        return "season removed";
+        return "Episode Removed";
       default:
         return a.type;
     }
   }
 
+  function toFullTitleCase(text: string) {
```
IRHM: The type of `text` can also be `undefined`, it shouldn't, but to avoid any errors, we could update the type to `string | undefined` and add a check to the top of the function to return `"unknown"` if the text is undefined.

src/routes/(app)/movie/[id]/+page.svelte
```diff
@@ -229,7 +229,7 @@
       <SimilarContent type="movie" similar={movie.similar} />
 
       {#if wListItem}
-        <Activity activity={wListItem?.activity} />
+        <Activity wListId={wListItem.id} activity={wListItem.activity} />
```
Comment (IRHM) [inline src/routes/(app)/movie/[id]/+page.svelte:232]: Also, we should add the new wListId prop to the <Activity/> components in tv and game pages. Currently the activity client state is only updated for tv.

src/lib/ActivityEditor.svelte
```diff
@@ -0,0 +1,98 @@
...
+
+<Modal title="Edit Activity" desc={activityMessage} maxWidth="400px" {onClose}>
+  <div class="centered">
+    <h3>Date</h3>
+    <input id="activity-date" type="date" bind:value={selectedDateString} />
+    <h3>Time</h3>
+    <input id="activity-time" type="time" bind:value={selectedTimeString} />
+
+    <div class="button-row">
+      <button class="delete" on:click={remove}>Delete</button>
```
IRHM: We have a `.danger` class for buttons that we could use here instead to keep the styles uniform.

src/lib/ActivityEditor.svelte
```diff
@@ -0,0 +1,98 @@
...
+      flex-flow: row;
+      justify-content: space-between;
+      margin-top: 10px;
+
+      button {
+        margin-top: auto;
+        width: max-content;
+      }
+
+      .delete {
```
IRHM: If using `.danger` class, we could remove this.

src/vars.scss
IRHM, line 1: If using `.danger` class, we could remove the delete-color variable.

server/activity.go
```diff
@@ -76,3 +80,31 @@ func addActivity(db *gorm.DB, userId uint, ar ActivityAddRequest) (Activity, err
...
+		return errors.New("id must be set to update an activity")
+	}
+	if activityUpdateRequest.CustomDate.IsZero() {
+		return errors.New("customDate must be set to update an activity")
+	}
+	res := db.Model(&Activity{}).Where("user_id = ? AND id = ?", userId, id).Update("custom_date", activityUpdateRequest.CustomDate)
+	if res.Error != nil {
+		slog.Error("Error updating activity in database", "error", res.Error.Error())
+		return errors.New("failed updating activity in database")
+	}
```
IRHM, line 95: We could add a check here to ensure `res.RowsAffected` is bigger than 0. This will ensure we return an error if we didn't actually update anything (return "activity not found" error or something like that).

Currently we return a 200 OK if we try updating activity we don't own or doesn't exist.

Same for deleteActivity.

server/routes.go
```diff
@@ -497,6 +497,41 @@ func (b *BaseRouter) addActivityRoutes() {
 		}
 		c.AbortWithStatusJSON(http.StatusBadRequest, ErrorResponse{Error: err.Error()})
 	})
+
+	activity.PUT(":id", func(context *gin.Context) {
```
IRHM: We might want to rename `context` to `c` to keep it uniform with the rest of the handler funcs.

Same for delete below.

server/routes.go
```diff
@@ -497,6 +497,41 @@ func (b *BaseRouter) addActivityRoutes() {
...
+		if err == nil {
+			err = updateActivity(b.db, userId, uint(id), activityUpdateRequest)
+			if err != nil {
+				context.JSON(http.StatusForbidden, ErrorResponse{Error: err.Error()})
+				return
+			}
+			context.Status(http.StatusOK)
+			return
+		}
+	})
```
IRHM, line 520: After the `err == nil` check, we need to return an error similar to we do above:

```
c.AbortWithStatusJSON(http.StatusBadRequest, ErrorResponse{Error: err.Error()})
```

server/routes.go
```diff
@@ -497,6 +497,41 @@ func (b *BaseRouter) addActivityRoutes() {
 		}
 		c.AbortWithStatusJSON(http.StatusBadRequest, ErrorResponse{Error: err.Error()})
 	})
+
+	activity.PUT(":id", func(context *gin.Context) {
+		userId := context.MustGet("userId").(uint)
+		id, err := strconv.ParseUint(context.Param("id"), 10, 32)
+		if err != nil {
+			context.Status(400)
```
IRHM: In places where we just return a 400 status (no error msg returned to user), it may be beneficial to add an error log.

If we ever got an unexpected 400 in the future, it would be a big help to see the error in the server log.

src/lib/Activity.svelte
```diff
@@ -187,12 +213,15 @@
       }
 
       li {
-        display: flex;
-        flex-flow: row;
-        align-items: center;
-        gap: 8px;
-        width: max-content;
-        max-width: 100%;
+        button {
```
IRHM, line 219: We could add `cursor: pointer` to this button so it's more obvious it's clickable.

IRHM (changes requested):
Thanks for looking at the comments I made before.

A lot of small comments, of course no rush to get them done, feel free to call me a poop head and I can help out with the changes if you'd like.

Apart from the comments, it's looking and working good so far!

stephaje:
All your comments looks reasonable to me. I'll work on changes sometime in the next ~24h

server/routes.go
stephaje: I don't agree on this one. I think the others should change. There is a very small benefit and a lot of drawbacks to using shortened variable names. Robert C. Martin has an entire chapter on naming in [Clean Code](https://www.goodreads.com/book/show/3735293-clean-code) which I believe to be the best take on the subject.

I recognize this is counter to [go's own recommendations](https://go.dev/talks/2014/names.slide#6), I just don't agree with them.

I'll go either way on it, just let me know.

IRHM: @stephaje I personally prefer keeping it short as long as its still quickly understandable (unfortunately dont have 30 buckaroos for Robertos book to see his points).

But regardless of recommendations, in this case, it's probably for the best to stick with the shorter variant `c` so it's inline with the other handlers, instead of refactoring the whole file (routes.go is a chubster).

IRHM (dismissed):
Very happy with the changes, thanks again for doing this!

There's one more small bug I've just noticed. If I type an invalid date and click update, theres an error logged in devtools console (`Uncaught RangeError: invalid date at api.ts:174:23`), and the Updating notification stays there indefinitely.

image

I'll let you decide if you'd like to add something to handle that, I'm happy to merge as is if not and add something a bit later.

_(I think it should let you merge this after i approve)_

stephaje:
Yeah, I'll take a look at it later this week. I borked my dev environment so I need some time to get things back together.

stephaje:
Hmm, I actually can't reproduce that. It just lets me set it without error.

image

If you want to write a follow up issue I'm fine with that, but I'd need more info to diagnose. I've added a little bit of client validation in the meantime, which can be bypassed, but if people really want to break the application like that, I say let them.

IRHM (approved):
@stephaje I agree with you. Funny it lets you set the year 20, I can't explain why my browser breaks.

Anyways good work! All looks great, I'm happy to get it merged.

PR #379, opened 2024-02-28
API keys support on the backend

### LINKED ISSUE #360: API Keys for users - giving other apps access to users data securely
It should be possible to access a user's data programmatically. This would allow for up to date data to be used by other systems such as recommenders.

titanventura:
Fixes #360 partially

### Changes made

- Added a table for API Keys
- added routes for users to be able to manage their API keys
- added a middleware to authenticate routes using API keys
- use the middleware for GET method of `/api/watched` endpoint

### DIFF
--- a/server/apikeys.go
+++ b/server/apikeys.go
@@ -1,0 +1,138 @@
+package main
+
+import (
+	"errors"
+	"log/slog"
+	"net/http"
+	"time"
+
+	"github.com/gin-gonic/gin"
+	"gorm.io/gorm"
+)
+
+type ApiKey struct {
+	ID        uint      `gorm:"primarykey" json:"id"`
+	CreatedAt time.Time `json:"createdAt"`
+	Value     string    `gorm:"not null" json:"value"`
+	UserID    uint      `gorm:"not null" json:"-"`
+	Note      string    `gorm:"not null" json:"note"`
+}
+
+type ApiKeyRequest struct {
+	Note string `json:"note,required"`
+}
+
+func (b *BaseRouter) addAPIKeyRoutes() {
+	apiKeyGrp := b.rg.Group("/apikeys").Use(AuthRequired(nil))
+
+	apiKeyGrp.POST("", func(ctx *gin.Context) {
+		var req ApiKeyRequest
+		if err := ctx.ShouldBindJSON(&req); err != nil {
+			ctx.AbortWithStatusJSON(http.StatusBadRequest, gin.H{
+				"error": "'note' field is required",
+			})
+			return
+		}
+
+		userId := ctx.MustGet("userId").(uint)
+		token, err := generateString(32)
+		if err != nil {
+			ctx.AbortWithStatusJSON(http.StatusInternalServerError, gin.H{
+				"error": "unable to generate an API key",
+			})
+			return
+		}
+
+		apiKey := ApiKey{
+			Value:  token,
+			UserID: userId,
+			Note:   req.Note,
+		}
+
+		if res := b.db.Create(&apiKey); res.Error != nil {
+			ctx.AbortWithStatusJSON(http.StatusInternalServerError, gin.H{
+				"error": "unable to add API key",
+			})
+			return
+		}
+
+		ctx.JSON(http.StatusOK, apiKey)
+	})
+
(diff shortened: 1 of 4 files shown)


server/apikeys.go
```diff
@@ -0,0 +1,138 @@
...
+type ApiKey struct {
+	ID        uint      `gorm:"primarykey" json:"id"`
+	CreatedAt time.Time `json:"createdAt"`
+	Value     string    `gorm:"not null" json:"value"`
+	UserID    uint      `gorm:"not null" json:"-"`
+	Note      string    `gorm:"not null" json:"note"`
+}
+
+type ApiKeyRequest struct {
+	Note string `json:"note,required"`
```
IRHM: I'm getting this warning: `unknown JSON option "required" (SA5008)` from go-staticcheck

server/apikeys.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func (b *BaseRouter) addAPIKeyRoutes() {
```
IRHM: We should move this to the `routes.go` file.

server/apikeys.go
```diff
@@ -0,0 +1,138 @@
...
+	apiKeyGrp := b.rg.Group("/apikeys").Use(AuthRequired(nil))
```
IRHM: I think `/apikeys` here could be `/api_keys`. Just to keep it inline with the naming of other routes.

server/apikeys.go
```diff
@@ -0,0 +1,138 @@
...
+
+	apiKeyGrp.POST("", func(ctx *gin.Context) {
+		var req ApiKeyRequest
+		if err := ctx.ShouldBindJSON(&req); err != nil {
+			ctx.AbortWithStatusJSON(http.StatusBadRequest, gin.H{
```
IRHM: We have an `ErrorResponse` struct we could use here instead (and in the other handlers), which would keep this a little more standardized with the rest of the handlers:

Example (from routes.go):
```
c.AbortWithStatusJSON(http.StatusBadRequest, ErrorResponse{Error: err.Error()})
```

server/apikeys.go
```diff
@@ -0,0 +1,138 @@
...
+
+	"github.com/gin-gonic/gin"
+	"gorm.io/gorm"
+)
+
+type ApiKey struct {
+	ID        uint      `gorm:"primarykey" json:"id"`
+	CreatedAt time.Time `json:"createdAt"`
+	Value     string    `gorm:"not null" json:"value"`
+	UserID    uint      `gorm:"not null" json:"-"`
```
IRHM, line 17: We may want to make this a foreign key that maps to a User to ensure data integrity.

server/apikeys.go
```diff
@@ -0,0 +1,138 @@
...
+	"errors"
+	"log/slog"
+	"net/http"
+	"time"
+
+	"github.com/gin-gonic/gin"
+	"gorm.io/gorm"
+)
+
+type ApiKey struct {
```
IRHM, line 13: An `UpdatedAt` field might be worth adding in here too.

IRHM (changes requested):
Hey @titanventura, thanks for making a PR (and sorry for taking so long getting to it)!

I had a quick look and noticed some changes we could make, let me know if you have any thoughts on the feedback.

I noticed the `APIKeyMiddlewareWrapper`, did adding the header check to the `AuthRequired` middleware not work out? I think that way would be preferable, just need to move the `x-api-key` header check into `AuthRequired` and if we find it, we can process it as an API key instead (in APIKeyMiddlewareWrapper, if you'd like, but probably renamed).

I know this is a draft, so you may have been getting to this, but the route handlers seem to have all the logic in them. We could separate most of it out to 'service' functions.

Thanks again for taking the time!

titanventura:
@IRHM , please go through the updated PR. Thanks !

IRHM (dismissed):

IRHM:
Hi @titanventura, I'm happy to get this merged. Thanks for the good work!

There are a few minor things, but It's probably best I do them after deciding how the UI will work so it isn't wasted effort incase it needs changing.

Thanks!

shummo:
Hi. How about this?

PR #387, opened 2024-03-03
Issue #358 - Plex Oauth

### LINKED ISSUE #358: Authentication: Add support for Plex/Jellyfin
It would be nice to allow users to log in using their Plex or Jellyfin account, to cut down on the number of credentials users need to manage. This can also lead to support for additional features down the road.

This is something that is supported by applications like [Jellyseerr](https://github.com/Fallenbagel/jellyseerr) and should be fairly straightforward to replicate.

stephaje:
### Changes made

Allow users to log in via Plex. Their username will be the same as their Plex username.

Some stuff to iron out, but just wanted to open this up for feedback/review.


src/lib/util/api.ts
```diff
@@ -337,6 +338,16 @@ export async function unfollowUser(id: number) {
     });
 }
 
+// TODO
```
stephaje: Hoping for some help specifically on this part, @IRHM. We need a server-unique but constant identifier here, and a way to fetch the current version. I was hoping you'd have thoughts on how this can be done.

server/auth.go
```diff
@@ -274,19 +291,68 @@ func register(ur *UserRegisterRequest, initialPerm int, db *gorm.DB) (AuthRespon
...
+	if err != nil {
+		return AuthResponse{}, err
+	}
+	slog.Info("Fetching metadata from Plex.")
+	username, err := fetchPlexUsernameFromToken(token.AuthToken)
+	if err != nil {
+		slog.Error("registerFirstPlexUser: Could not fetch Plex username", "error", err)
+		return AuthResponse{}, errors.New("could not fetch plex username")
+	}
+	user := UserRegisterRequest{Username: username, Password: token.AuthToken, Type: PLEX_USER}
```
stephaje: Using the auth token as the password is just temporary. In the future, when pulling plex history, the token would need to be stored somewhere in plaintext, and the password should instead be a non-related random string. This works for the purposes of only supporting oauth log in, though.

IRHM:
Hi @stephaje, thanks for working on this. Sorry I've had a busy week, will reply to the comments now.

src/lib/util/api.ts
IRHM: @stephaje If the clientIdentifier must be unique for each selfhosted Watcharr instance, then I suppose we could generate one and place it in the watcharr.json config (similar to how we generate the JWT_SECRET in the config) when needed (when plex auth is enabled by an admin). If it only needs to be unique for Watcharr and doesn't matter if different instances share it, we could just hard code a string here.

We have a global variable defined in Vites config: `__WATCHARR_VERSION__` which we use in the face menu, that might work for the version here. If it doesn't, I'm sure it doesn't matter if we just hard code that property at `"1.0.0"`.

server/auth.go
IRHM: @stephaje I'm a little confused, but could it work like jellyfin auth does? I assume every time the plex user wants to login, they would have to go through the same oauth procedure with plex.

If that will give us a token and some sort of unique identifier, we could store them in the `third_party_id` and `third_party_auth` fields. Then on each login, we just compare the users id we get from plex to our `third_party_id` (if none exists, we can create them a user).

Then the password field can be left empty (like we do with jellyfin users), since we won't use it, and we will have the auth token stored for later use.

Edit: I noticed there's register and login with plex buttons, if we just have the login with plex button and create a user if they don't exist locally, that could work better (like mentioned above)?

src/lib/util/api.ts
stephaje: > (when plex auth is enabled by an admin)

I think it needs to be done immediately so that the admin can use oauth for their own log in. Then it can be disabled if they don't want to use it. The only concern I see here is the possibility that if OAuth breaks or goes down, the admin could be locked out of the account.

Maybe I should write it to support both a base login + associated Plex account?

Thanks for the code pointers.

IRHM: No problem @stephaje,

I guess thats just the downside of using Plex, if their servers go offline, theres not much you can do.

For now, maybe the first admin account should be a normal user, like it is currently. Then they can login via plex to make a new account and give it admin if they'd like.

Then hopefully in the future when we allow linking services to your account we can make this easier for admins.

The confusing part for allowing plex auth and normal auth as backup, is that usernames might not be unique for normal users and plex ones.

If first admin must be a normal user, at least the login can be saved for an emergency. (Then a plex one can be made and given admin if wanted).

server/auth.go
stephaje: Yeah, I can take a look at doing it that way instead. I wasn't sure if the API's were still protected w/o the password filled out and I didn't get to reading the code for it just yet.

src/lib/util/api.ts
stephaje: I'm okay with that strategy. I'll take a look at updating it later today.

server/auth.go
IRHM: That's awesome, thanks @stephaje!

Yep as long as we only give the user their watcharr token after we authenticate them against plex (and locate/add them to our db) then we're all good

stephaje:
@IRHM The only comment I didn't accommodate was merging the login and register paths, just because I think it's better to keep the separate to more cleanly support when new signups are disabled.

Everything else should be more in line with the jellyfin experience.

I did separate the plex button from the other buttons because it makes more sense visually to not tie the plex button to the login and password fields, but feel free to tweak as desired.

I also added color to the Jellyfin button, but feel free to revert or pick a more suitable color.

IRHM (approved):
Thanks for the great work @stephaje!

Issue #359, opened 2024-02-16
Importing: Support syncing Watch History and Ratings from Plex Login


stephaje:
If a user is signed in with a Plex account, they should have the option to sync their watch history and/or their media ratings. This mildly overlaps with https://github.com/sbondCo/Watcharr/issues/82.

IRHM:
Hey @stephaje, once #358 and #82 are completed I will definitely come back to this.

Hopefully the code will be mostly reusable for syncing so after #82 this should be much simpler.

stephaje:
@IRHM I can start taking a look at this, unless you'd prefer to.

IRHM:
> @IRHM I can start taking a look at this, unless you'd prefer to.

Hey, feel free, I'll assign you. Thanks!

Should be able to use `jellyfin_sync.go` as a rough guide for getting it working for plex. Sorry in advance for the ungodly size of the main function in there, I didn't think it was worth (the time) separating out something like that.

PR #401, opened 2024-03-10
Issue #359: Importing: Support syncing Watch History and Ratings from Plex

stephaje:
### Changes made

- Generalize sync modal to work for both plex and jellyfin
- Fix plex login bug due to bad type filtering
- Update plex token on re-login to support token revoke/expiration
- Add Plex server configuration for fetching watch data
- Add TV and Movie specific TMBD APIs
- Move activity modal into modals directory

### DIFF
--- a/server/plex_sync.go
+++ b/server/plex_sync.go
@@ -1,0 +1,149 @@
+package main
+
+import (
+	"errors"
+	"log/slog"
+	"time"
+
+	"gorm.io/gorm"
+)
+
+type PlexSyncResponse struct {
+	JobId string `json:"jobId"`
+}
+
+// Perform a Plex sync.
+// Errors are added silently to the job.
+func plexSyncWatched(
+	db *gorm.DB,
+	jobId string,
+	userId uint,
+	userThirdPartyAuth string,
+) {
+	updateJobCurrentTask(jobId, userId, "fetching libraries")
+	libraries, err := plexGetLibraries(userThirdPartyAuth)
+	if err != nil {
+		addJobError(jobId, userId, "failed to get plex libraries")
+		updateJobStatus(jobId, userId, JOB_DONE)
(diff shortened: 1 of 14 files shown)


server/plex_sync.go
```diff
@@ -0,0 +1,149 @@
...
+				match := sr.Results[0]
+				_, err = addWatched(db, userId, WatchedAddRequest{
+					Status:      FINISHED,
+					ContentID:   match.ID,
+					ContentType: MOVIE,
+					Rating:      int8(movie.UserRating),
+					WatchedDate: time.Unix(int64(movie.LastViewedAt), 0),
+				}, IMPORTED_WATCHED_PLEX)
+				if err != nil {
+					if err.Error() == "content already on watched list" {
```
stephaje, line 70: I think here we should attempt to update the rating, thoughts @IRHM ?

IRHM: Good idea @stephaje, but I think that might cause problems where a user changes their rating on Watcharr, then a full re-sync happens with plex and their rating is overridden.

I suppose we could do a two-way sync with plex to get around that, but that sounds pretty complex. I would imagine most people only sync, then only interact with the show (for ratings/status) through watcharr (not sure)?

stephaje: I would imagine most people would resync every couple of weeks to pull in fresh watches. If they watched something but hadn't rated it when they did the first sync, then rated it in plex, then resync, it would be nice to pull in. Maybe it will update only if the current watcharr rating is empty?

IRHM: Sounds good @stephaje. As long as it can't accidentally override a rating set in watcharr it should be alright.

IRHM:
Ported changes to a fresh branch, closing this PR in favor for it in #413.

PR #413, opened 2024-03-16
Plex sync

### LINKED ISSUE #359: Importing: Support syncing Watch History and Ratings from Plex Login

IRHM:
### Changes made

Ported from #401 with some changes (easier than fixing old pr first).

Changes from old pr:

- Also sync seasons/episodes.
- Use `includeGuids` query parameter (that was buried in a forum post... why...) to get `tmdbId` so we can directly get a show/movie without needing to rely on search.
- Don't move activity modal component.
- Also some other small changes, but not worth mentioning because I forgot.

Closes #359


IRHM:
dumbo plex auth token doesnt work against the local server for api requests when using another account :((((((((((((((

but it does work if i login via web ui and steal the auth token it generates which makes me think something is wrong in the oauth flow and we aren't asking for a token with permissions for our local server??? plos help no i dont need help ill figure it out tomorrow but also help

IRHM:
Figured it out, but in the progress moved plex auth tokens to a new `user_services` table because we now have two separate tokens (one for use against plex.tv and another for our home server).

At least we now have a starting point when integrating future services/integrations.

Jellyfin can stay how it is for now (and forever probably) to avoid breaking changes.

PR #423, opened 2024-03-20
Configurable data location 

mordquist:
Running Watcharr as a service on NixOS requires changing the data directory path since the nix store (where the binary is located) is read-only.

### Changes made

I'm simply reading the path from an environment variable if set with a fallback to the default `./data` if not. I've also updated the places where the directory was hard-coded to use the `DataPath` var.

### DIFF
--- a/server/config.go
+++ b/server/config.go
@@ -6,13 +6,20 @@ import (
 	"log"
 	"log/slog"
 	"os"
+	"path"
 	"time"
 
 	"github.com/sbondCo/Watcharr/game"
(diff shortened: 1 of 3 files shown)


IRHM (approved):
Hey @mordquist, thank you very much for doing this and sharing!

I made a small change to fix a typo in the environment variables name, but other than that everything looks great!

I'm sure this will help out others in similar situations too.

mordquist:
Thank you for the kind words and the fast merge even though i can't spell the name of the project properly :)

PR #433, opened 2024-03-24
UI tweaks

reitenth:
image
image

Changes:

- smaller main poster to match other cast/similar size -> made posters universal height
- adjusted left padding to be universal from main poster/cast/similar/activity/seasons
- added harsher gradient for cast names

Want to Change:

- tmdb vote/episode vote to be inline with title/year and episode title/time
- more prominent rating on home screen, maybe like how ryot does it with top right overlay


IRHM:
Hey @reitenth, thanks for making these changes!

Good catch on the Activity section, it looks much better with your changes.

I'm not too sure on the main poster being shrunk to match the sizes of the other posters, I like how the main one is emphasized. It's also easier to see/appreciate the poster when it's a bit bigger. Maybe we could look into making a `compact` theme, if you prefer it that way?

Feel free to make the rating inline with the contents title, I'm sure it would look better that way.

We have a `detailed view` that shows the rating and status of a show on your watched list (toggled via the eye icon menu on navigation bar), is that the sort of prominent rating you were thinking of?

image

IRHM:
Hey @reitenth, I haven't heard back from you in a while so I'm going to close this PR for now.

Please feel free to open an issue with any features/ui related improvements you might have in the future, I'm happy to discuss anything.

Thanks again for opening this pull request!

Issue #452, opened 2024-03-31
[Suggestions] Various UI & Functionality Changes


LoganJFisher:
Below are a list of UI and functionality changes that I would like to recommend. They are not listed in any particular order.

- [ ] Different font for ratings (the current chalkboard font looks a bit out of place).
- [x] #453
- [ ] Hover left and right in "Discover" to scroll lists (rather than clicking and dragging the scroll bar).
- [ ] When on the "Discover" page, replace the "Discover" icon with a home icon.
- [ ] "Recommended" shows and movies on the "Discover" page. A random list of shows or movies that are listed as "similar" to shows and movies you've rated highly, removing any that you've already marked as "finished" or "watching". Generate a new list on page refresh.
- [ ] When marking a show's season as "finished", automatically mark all individual episodes of that season as "finished" (also all episodes if the entire show is marked as "finished").
- [ ] Automatically change shows marked as "finished" to "On Hold" when one or more new episodes release.
- [ ] Similarly, when the final episode of a show is marked as "finished", automatically mark the season and the entire show as "finished"
- [ ] Shows should provide a notification (a bell or envelope icon at the top) when new episodes are released. For shows with subbed and dubbed episodes (e.g. anime), you should be able to select which one you want to be notified for (or both). Related: #409 
- [ ] Filter content on watchlist by country of origin.
- [ ] #476
- [ ] Bulk selecting episodes for modifying status.

Apologies if any of these have already been requested elsewhere. Please don't take these as criticisms of Watcharr - I'm very happy with it as a whole, but think these changes could make it even better.

IRHM:
Hi @LoganJFisher, thanks for taking the time to write these suggestions out, I appreciate it!

Just two clarifications on the things I am unsure of:

- For the ratings font, do you mean on posters (a screenshot may help me understand, if possible)?
- For the ratings being refined to work with decimals, would it just be in increments of 0.5 or would you want to whole range to be accessible in increments of 0.1? I think I'd like to add different rating methods as a setting for users in the future (so everyone is happy).

The rest of the points sound good, I may split them out to other issues when I have the time to work on them. Thanks again!

LoganJFisher:
![SmartSelect_20240331_123416_Folder Widget.jpg](https://github.com/sbondCo/Watcharr/assets/63993409/3d91d162-6557-4116-9515-74928f4e2a88)

[...]

PR #463, opened 2024-04-03
Ability to set a country, both in server and user settings, to have the correct streaming provider (but doesn't affect language)

### LINKED ISSUE #117: Add user `region` setting
Caching won't work with the current design (https://github.com/sbondCo/Watcharr/issues/59#issuecomment-1688975877), this new flow, as described below will work and looks to be a better solution in general.

Currently the design is to:
- User makes request to endpoint
- `WhereaboutsRequired` middleware (should) gets the servers region and the users region. If the user has a `region` preference, then that will be used, if not, the servers default `region` will be used.
- TMDB request is carried out with the `region`

The new flow:
- When user logs on, we should get their region preference from their user settings.
- Add a GET query of `region` to every request to an endpoint that want a `region`.

The new flow is simpler and will work with the caching plugin. We also get the added benefit of not requiring a db call (for user region preference) at every endpoint call that needs it.

`WhereaboutsRequired` luckily hasn't been fleshed out yet, so we just need to remove it and continue with the new approach instead.

- [ ] ~Remove `WhereaboutsRequired` and use of the `userCountry` variable it sets (probably the whole middleware.go file too).~
- [x] Add user region preference setting for user.
- [x] Test caching to ensure it works and takes into account each region.

stignarnia:
I felt like this feature was needed, and I also found issues #452 and #117 mentioning it!

I'm very inexperienced and this is one of my first PRs in general, so I know what I did works but it might be stupid, have a wrong style or who knows what, so I'm sorry in advance and feel free to reject this or make any necessary changes.

#### You can choose the default country new users will be onboarded in
immagine

#### Users can choose the country for the streaming providers
immagine

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -425,6 +425,7 @@ func loginJellyfin(user *User, db *gorm.DB) (AuthResponse, error) {
 			dbUser.ThirdPartyAuth = resp.AccessToken
 			dbUser.Username = resp.User.Name
 			dbUser.Type = JELLYFIN_USER
+			dbUser.Country = &Config.DEFAULT_COUNTRY
 
 			dbRes = db.Create(&dbUser)
 			if dbRes.Error != nil {
@@ -487,6 +488,7 @@ func loginPlex(lr *PlexLoginRequest, db *gorm.DB) (AuthResponse, error) {
 				AuthToken:  lr.AuthToken,
 				AuthToken2: homeAuthToken,
 			})
+			dbUser.Country = &Config.DEFAULT_COUNTRY
 			dbRes = db.Create(&dbUser)
 			if dbRes.Error != nil {
 				slog.Error("loginPlex: Failed to create new user in db from plex response", "error", dbRes.Error)
--- a/server/auth.go
+++ b/server/auth.go
@@ -88,6 +88,8 @@ type UserSettings struct {
 	// even if the watched item state has since been changed.
 	// Also if user wants to show in watched stats.
 	IncludePreviouslyWatched *bool `gorm:"default:false" json:"includePreviouslyWatched"`
+	// User's country to get correct content streaming providers.
+	Country *string `gorm:"default:'US'" json:"country"`
 }
 
 // Holds third party service auth tokens for users.
--- a/server/auth.go
+++ b/server/auth.go
@@ -292,6 +292,8 @@ func register(ur *UserRegisterRequest, initialPerm int, db *gorm.DB) (AuthRespon
 		user.Permissions = initialPerm
 	}
 
+	user.Country = &Config.DEFAULT_COUNTRY
+
 	res := db.Create(&user)
 	if res.Error != nil {
 		// If error is because unique contraint failed.. user already exists
--- a/server/middleware.go
+++ b/server/middleware.go
@@ -4,22 +4,28 @@ import (
 	"log/slog"
 
 	"github.com/gin-gonic/gin"
-	"gorm.io/gorm"
 )
 
 // Location middleware
-func WhereaboutsRequired(db *gorm.DB) gin.HandlerFunc {
+func WhereaboutsRequired() gin.HandlerFunc {
 	return func(c *gin.Context) {
-		userId := c.MustGet("userId").(uint)
-		settings, err := userGetSettings(db, userId)
-
-		if err != nil {
-			slog.Error("Failed to get user settings", "error", err.Error())
+		region := c.Query("region")
+		slog.Debug("WhereaboutsRequired: middleware hit", "region", region)
+		if region == "" {
+			// If no region is passed, default to server region.
+			if Config.DEFAULT_COUNTRY != "" {
+				slog.Debug("WhereaboutsRequired: Using server default country.", "default_country", Config.DEFAULT_COUNTRY)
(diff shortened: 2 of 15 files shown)


IRHM:
Hey @stignarnia, I apologize for it taking so long before I could look at this PR. I really appreciate your great work!

I did make a few modifications (some I've had planned in my head for a while), hopefully that's ok! I thought it would be a little easier than asking for a million changes from you.

So that you understand my reasoning, I'll add a note on each change I've made (hopefully they make sense and can help).

Thank you again!

server/auth.go
```diff
@@ -421,6 +425,7 @@ func loginJellyfin(user *User, db *gorm.DB) (AuthResponse, error) {
 			dbUser.ThirdPartyAuth = resp.AccessToken
 			dbUser.Username = resp.User.Name
 			dbUser.Type = JELLYFIN_USER
+			dbUser.Country = &Config.DEFAULT_COUNTRY
```
IRHM, line 428: The `loginJellyfin` and `loginPlex` methods will create a user if they don't exist in our database, so added the default country for them too.

server/middleware.go
```diff
@@ -8,13 +8,24 @@ import (
 
 // Location middleware
 func WhereaboutsRequired() gin.HandlerFunc {
```
IRHM, line 10: I changed this around to look for a `query` parameter on the request (I had planned to do this a little while ago #117).

The issue explains in more detail, but the tl;dr is it saves us making an extra database call and means we don't need to flush cache each time the user changes their settings.

server/routes.go
```diff
@@ -122,7 +122,7 @@ func (b *BaseRouter) addContentRoutes() {
 	}))
 
 	// Get tv details (for tv page)
-	content.Use(WhereaboutsRequired()).GET("/tv/:id", cache.CachePage(b.ms, exp, func(c *gin.Context) {
+	content.GET("/tv/:id", WhereaboutsRequired(), cache.CachePage(b.ms, exp, func(c *gin.Context) {
```
IRHM, line 125: Moved the usage of `WheraboutsRequired`. When adding it with `.Use`, the middleware will apply to all subsequently defined routes. 

Adding it as a handler before our main handler func means it will only apply to this route.

src/lib/DropDown.svelte
```diff
@@ -11,15 +14,54 @@
...
+    let f: string | number | undefined;
+    if (filteredOptions.length > 0) {
+      f = typeof filteredOptions[0] === "string" ? filteredOptions[0] : filteredOptions[0].id;
+    }
+
+    // Find first button with text content starting with letter pressed and scroll it into view.
+    const btns = ulElement?.querySelectorAll("button");
+    for (let i = 0; i < btns.length; i++) {
+      const btn = btns[i];
+      if (btn.textContent?.toLowerCase()?.startsWith(pressedLetter)) {
```
IRHM, line 42: Main change, it'll scroll to the first button with text starting with the letterPressed.

Had to change so it would work with the new way the data is passed.

May have to change in the future, I think it'll only work well if the data in the dropdown is sorted alphabetically, which it isn't guaranteed to be (for now it is, with regions).

src/lib/DropDown.svelte
```diff
@@ -11,15 +14,54 @@
...
+    if (typeof active === "string" && !isDropDownItem) {
       activeValue = active;
     } else {
       const v = options.find((o) => (typeof o !== "string" ? o.id === active : false));
       if (v && typeof v !== "string") activeValue = v.value;
     }
   }
+
+  onMount(() => {
+    document.addEventListener("keypress", handleKeyPress);
```
IRHM: Moved this to onMount so it's only added once.

src/lib/DropDown.svelte
```diff
@@ -11,15 +14,54 @@
...
+
+    return () => {
```
IRHM, line 61: Added this, so when the component is destroyed (eg user navigates away to different page) we remove the event listener.

src/lib/RegionDropDown.svelte
IRHM, line 1: Created new RegionDropDown component so we could reuse the logic for getting regions.

server/routes.go
IRHM: I just realised, this might have been me that added it with `.Use` 💀

src/lib/DropDown.svelte
stignarnia: I'm not sure it is actually, I think it's sorted alphabetically by county code, but we display county names

stignarnia:
Thank you so much for taking a look at it! Ofc changes are welcomed, they all make sense except there may be a problem with the auto scroll one. Thank you for merging as well!

src/lib/DropDown.svelte
IRHM: @stignarnia Oh that makes sense. I changed the way the data is provided to the dropdown, so guess I had to change this anyways (I've forgot now, but I believe there was no way to get the button by it's full value anymore after my change so I decided to just rely on the first letter).

stignarnia: I can't say I understand what you did, but I tested it, and it works. For example, "VA" is correctly sorted among other countries starting with H as the English name is "Holy See".

PR #563, opened 2024-06-25
Ryot import implementation

### LINKED ISSUE #556: [Feature] Import from Ryot
Hi :v:

I've been testing watcharr lately and it seems to be great!
However, I'm trying to migrate from [Ryot](https://github.com/IgnisDa/ryot) and the current import methods don't support that.
I have a bit of experience with JS and am fairly confident I can implement this new import option.
As far as I understand, this would be done by adding a new else if entry inside the "getList()" function in "/src/routes/(app)/import/process/+page.svelte"

Is this something you guys are interested in?
If so, any tips would be much appreciated :+1:

### LINKED ISSUE #565: [Enhancement] Better units
Hi :v:,

I'd like to propose some changes regarding the time units used.

1. Abbreviation of minutes as "min" both in series and episodes summaries:

      image  => image

      image => image

2. In profile/settings, change the total count to months/weeks/day/hours instead of hours/min:

      image => image

If these ideas are in line with the project's vision , I make myself available to implement them

oPisiti:
### Changes made

1. Created a new button on /import/ page for Ryot
2. Added Ryot svg logo
3. Created processRyotFile() function to correctly import the data of the json ryot export


IRHM:
Hey @oPisiti, I glanced over the changes and they look great, thank you for implementing this!

I will hopefully get to test the changes tonight or tomorrow, then I can get this merged.

Thanks again!

IRHM:
Hey @oPisiti, looks like some of the code wasn't formatted, I can format it for you if you'd like (feel free to do so yourself if you want to, we use prettier, running `npm run format` should format everything automatically)?

Also, if it would be alright, do you think you could attach a ryot export example that I could test with? I tried using their demo instance to export, but it doesn't seem to be working.

oPisiti:
Hi @IRHM,

Below is the custom export I tailored to test the import function.
It contains:
- Movies with different and multiple categories, in order to test the status ranked on line 425 of src/routes/(app)/import/+page.svelte. This rank is important to define the most meaningful tag, as Ryot allows multiple.
- Movies with reviews
- Series with multiple episodes watched and different progress levels
- Series with no tags (which means dropped)
- Series with episode reviews, which are not supported in watcharr (should be dropped without errors)
- Other types of media, which should all raise errors: anime, manga, audiobook, visual novel

P.S.: In Ryot, there is overlap between anime and tv shows. As the anime section yields weird results when searching, I only used the tv category for my watched anime and it worked flawlessly and I imagine most people do the same. Also, it uses anilist identifiers. Therefore, I disconsidered "anime" as a source

[RyotMedia - min.json](https://github.com/user-attachments/files/16022149/RyotMedia.-.min.json)

IRHM:
Hey @oPisiti, thank you very much for doing this, I just made some small tweaks, but other than that I am very happy to get this merged in!

I mostly just fixed some type errors, I will add comments to the code in the commit so you can review the changed I've made.

Kind of irrelevant to what this PR was for, but I also changed the Stats styling to push the values down instead:
**From**
image
**To**
image

I think that looks nicer than what we had before?

Thanks again!

src/routes/(app)/import/+page.svelte
```diff
@@ -406,7 +407,7 @@
       // Build toImport array
       const toImport: ImportedList[] = [];
       const fileText = await readFile(new FileReader(), file);
-      const jsonData = JSON.parse(fileText)["media"] as Watched[];
+      const jsonData = JSON.parse(fileText)["media"] as any[];
```
IRHM, line 410: Changed to `any` array, the data in the ryot export didnt conform to the `Watched[]` type

src/routes/(app)/import/+page.svelte
```diff
@@ -422,7 +423,7 @@
 
         // Define the main general status of the movie/show
         // In Ryot, it can be marked as multiple of the following, so choose the most relevant
-        const statusRanks = [
+        const statusRanks: [string, WatchedStatus][] = [
```
IRHM, line 426: Gave this a "tuple array" type so that it knows the second value is a `WatchedStatus`, fixed a type error below where we add `status` to the `ImportedList` object.

src/routes/(app)/import/+page.svelte
```diff
@@ -448,23 +449,23 @@
           thoughts: v.lot === "movie" && v.reviews.length ? v.reviews[0].review.text : "",
 
           // Ryot does not support overall rating for shows
-          rating: v.lot === "movie" && v.reviews.length ? Number(v.reviews[0].rating) : null,
+          rating: v.lot === "movie" && v.reviews.length ? Number(v.reviews[0].rating) : undefined,
```
IRHM, line 452: Changed `null` usage to `undefined` to conform to the `rating` property type.

**Irrelevant, but maybe informative if you care:**
I usually try to avoid using `null` and use `undefined` instead. Since JS has null and undefined, they usually tell us the same thing, so I prefer to, in most cases, strictly use `undefined` instead (can't think of any upsides to `null` other than if our logic needs to differentiate between the two for some reason). Anyways that's just what I think.

src/routes/(app)/profile/+page.svelte
```diff
@@ -145,14 +145,14 @@
...
+    let ansString = "";
     let tmp;
     for (const c of countInMinutes) {
-      tmp = Math.floor(m / c[1]);
+      tmp = Math.floor(m / (c[1] as number));
 
       // Ignore fields with fewer than 1 unit
       if (tmp) ansString += `${tmp} ${c[0]}${tmp >= 2 ? "s, " : ", "}`;
-      m -= tmp * c[1];
+      m -= tmp * (c[1] as number);
```
IRHM, line 155: Casted these to be numbers, I could've done the tuple array thing here too but i didn't think of that until now and its too late!!!!!! AAAAAAA

src/types.ts
```diff
@@ -34,6 +34,7 @@ export type Icon =
   | "eye"
   | "star"
   | "movary"
+  | "ryot"
```
IRHM, line 37: Added `ryot` to the `Icon` type (every icons name is included in this type).

src/routes/(app)/import/+page.svelte
oPisiti: Oh yeah, that makes sense.

JS is not my main language, so I was going with knowledge from a few years ago when I decided on null.
I will keep that in mind for future PRs

Issue #461, opened 2024-04-02
[Feature Request] OAuth authentication


Steeven9:
The docs mention that you can log in with Jellyfin, but why not take this one step further and allow to register your own OAuth 2.0 provider for SSO?
For example, one might have a local Keycloak instance set up, or want to use their Google account.

IRHM:
Hey @Steeven9, thanks for opening this issue!

We can definitely look into this at some point, there's quite a bit of other tasks ahead of this, so I'm not sure when I'll be able to look at it. If you or someone else would like to, feel free!

I have never hosted/used my own instance of a oauth/sso provider, hopefully they are all the same so we can just add generic setting that you could use for any service.

Steeven9:
Thanks for the quick reply! I unfortunately have never implemented an SSO login, only _configured_ a few so far :)

Thankfully OAuth/OIDC is a pretty vastly used and well-documented standard so it shouldn't be too hard to find some libraries/examples to implement it, for example https://authjs.dev

mgrimace:
I'm jumping in to speak in support of this, specifically for [Authentik](https://goauthentik.io) support. Ideally, to allow reverse proxy authentication by header. 

I have setup NPM proxy, with Authentik for authentication sitting before Watcharr and other services. So every time I open Watcharr's webui, NPM redirects it to Authentik for authentication. Authentik can use Plex as an SSO, and would ideally pass on the user and other related proxy authentication headers to Watcharr. If there is already a user with the same username, Watcharr would then automatically login in the user without requiring it's own login screen.

The benefit of this is that I can use secure athentication service, and I wouldn't have to login twice - if that makes sense.

AFAIK Authentik uses the header `X-authentik-username`, which is how I accomplish this in Calibre-web.

lufixSch:
I would also really appreciate OAuth/OIDC support, but I understand this might be more complicated (I also haven't implemented it before)
An alternative that could be forward auth as mentioned by @mgrimace. This should be as far as I understand relatively straight forward to implement.

royaltongue:
Is this still being considered? I'd also appreciate being able to have my family log in with OIDC through Authentik

lufixSch:
@royaltongue Forward Auth/Proxy Authentication is implemented and works well for me using Authentik with Traefik as a reverse proxy.

OIDC is not implemented.

mgrimace:
[...]

PR #632, opened 2024-09-18
Proxy authentication flow

lufixSch:
This PR aims to add forward auth/proxy auth to this project as discussed in #461. In forward auth a proxy routes requests through a third party authentication service and only if this service returns a 200 response, the request is passed to the actual service. The authentication service will add one or many authentication headers to the request depending on the service which can be used to identify the authenticated user.

### Changes made

In an ideal implementation the user would be authenticated automatically (e.g. no 'login' button has to be pressed) but with the current implementation I wasn't able to think of a viable solution which doesn't require a DB lookup on every API request.
Instead, I added another provider ("proxy") and a corresponding login button in the frontend. When this button is pressed an API request to `/auth/proxy` is made which checks for the authentication header and returns an authentication token if the header is valid.

The current implementation is a minimal proof of concept in order to see if it could work. It's the first time I worked with GO, and I'm open to feedback.

### Missing features

- [x] Settings for enabling proxy auth and specifying the name of the authentication header
- [x] Allow creation of a new user if it doesn't already exist
- [x] Protect `/auth/proxy` route so that it only works if proxy auth is specifically enabled


lufixSch:
I added automatic user creation. The current behavior is the following:
If the user already exists either with no `UserType` (e.g. username + password) or `PROXY_USER` they are authenticated. 
If no user exists with that username and either no `UserType` or `PROXY_USER` as type then a new user with `UserType` `PROXY_USER` is created.
A user of type `PROXY_USER` can not log in with username and password.

lufixSch:
The authentication header can now be specified using the `watcharr.json` config file or through the admin settings in the frontend

The `/auth/proxy` route will return `403` as long as no authentication header is specified in the config. It also returns `403` if the authentication header is missing or if it is empty (e.g. "")

christaikobo:
@IRHM Can you please take a look at this? It's really useful, thanks!

IRHM:
Oh sorry! I hadn't realised this was completed. Will hopefully have a look for the next release.

Would someone be able to list some common proxies that people use? I would like to try them out to test this pr.

I think authentik and authelia?

christaikobo:
Usually people use caddy, traefik or nginx as proxy, authelia or authentik
as forward authentication service.

IRHM:
Thanks for confirming @terrytw, I appreciate that! I will look at testing with both when I find the time. Hopefully that covers most people who'd like to use this functionality.

lufixSch:
I'm using traefik together with authentik. 
I've been using this MR for the past months now and had no issues.

server/auth.go
```diff
@@ -374,6 +375,38 @@ func login(user *User, db *gorm.DB) (AuthResponse, error) {
 	return AuthResponse{Token: token}, nil
 }
 
+func loginProxy(user *User, db *gorm.DB) (AuthResponse, error) {
+	slog.Debug("A User Is Logging In", "username", user.Username)
+	dbUser := new(User)
+	res := db.Where("username = ? AND (type IS NULL OR type = 0 OR type = ?)", user.Username, PROXY_USER).Take(&dbUser)
```
IRHM, line 381: I think this query should only look for `type = ?` so that we never allow logging into a non proxy user (type=3).

IRHM (changes requested):
Thank you for doing this @lufixSch - i found one issue at the moment (i can do the change if you'd like).

I did want to ask if you know if other apps/servers handle forward auth/sso like this (standard practice)? If I am correct, it seems your code looks for a header (PROXY_AUTH_HEADER) containing a username and if found, allows login (this is foreign to me, forgive me if I am being dumb). In my head it seems scary, but I suppose if setup correctly (watcharr only accessible through proxy & forward auth) it should be safe.

I did setup Authelia locally to test this and it does seem to work well.

To mention your note about supporting automatic auth, I think we could add extra code to the login page to scan available providers returned from the `/auth/available` call (kind of like how we check if its there when we show the SSO button), if we see `proxy` we could run your `proxyLogin()` function without any user input (we could also make a new setting PROXY_AUTO_LOGIN if you think that makes sense).

christaikobo:
> If I am correct, it seems your code looks for a header (PROXY_AUTH_HEADER) containing a username and if found, allows login.

This is pretty standard, I self host several services myself and it's usually how it's handled. It is dangerous and the responsibility falls on the users to correctly configure it. 

Of course full OIDC is better but it's so much more work, and proxy header auth is a decent compromise. 

One thing I would like to remind you guys is that with this kind of authentication, maybe it's better to either disable or hide logout button. Maybe it's already in the code and I'm blind and didn't see it, in that case, my apologies.

IRHM:
> This is pretty standard, I self host several services myself and it's usually how it's handled. It is dangerous and the responsibility falls on the users to correctly configure it.

Thanks for your input @terrytw, it does make me feel better knowing you agree and that its common.

> One thing I would like to remind you guys is that with this kind of authentication, maybe it's better to either disable or hide logout button. Maybe it's already in the code and I'm blind and didn't see it, in that case, my apologies.

I tried looking up why apps may disable the logout button when forward auth headers are used, but couldn't find much (likely because i'm using the wrong key terms).

Anywho, I believe the current code issues a (watcharr) JWT token to the user after successful SSO/proxy login, so the logout button does do stuff in our case and could allow users to login to non sso accounts if they desired? (Our logout button doesn't perform any http requests, if i remember correctly, it just simply clears all app state in the client)

I wonder if you think it makes sense for us to keep it in our case, since it still performs a function? Maybe we could rename it for proxy users if you think that could make sense/be less confusing.

lufixSch:
@IRHM As @terrytw already said, this is a pretty common method for easy authentication with a shared account in a self-hosted environment. I also agree, that OIDC would be preferable. However it is far more complicated to implement.

You are also right, with your assessment, that this poses a significant security risk, if the setting is enabled, when Watcharr is **NOT** running behind a proxy. For this reason I added the warning "Only set this if Watcharr is running behind a trusted proxy" on the settings page.
I could emphasize this (bold/red text or similar) to make sure no one enables this by accident if you want.

Your Idea with the automatic login sound good. It would be nice if you could implement it. I can also do it but as I am pretty busy right now it would probably take until March.

@terrytw Concerning the logout button I agree with @IRHM it still needs to be enabled. With the current login solution Watcharr only checks for the proxy header on login and uses the JWT afterward. So if you want to change users you would still need to press the logout button.

IRHM:
Thanks for your comments @lufixSch, it makes a lot of sense!

and thanks again for doing the heavy lifting on this, I can look into those smaller changes for us shortly.

server/auth.go
lufixSch: I solved it this way to allow the following behaviour I already explained above:

> If the user already exists either with no UserType (e.g. username + password) or PROXY_USER they are authenticated.
> If no user exists with that username and either no UserType or PROXY_USER as type then a new user with UserType PROXY_USER is created.

With this method you can migrate an existing setup to forward authentication without issues (as long usernames in watcharr and your auth solution match) and the Admin user created when first launching Watcharr can also be migrated to forward auth.

If you don't like this behavior I can of course change it. But as soon as you implement automatic auth with forward auth, it is needed as otherwise you won't be able to log into the main account anymore.

IRHM: Ah okay I see @lufixSch, that does make more sense, I can understand server admins wanting an easy migration for using their existing accounts with forward auth.

I think it might be possible to break, for example, if someone logs in via proxy and they have a PROXY_USER type, but then they or someone else makes a normal user type with the same username. When the user tries logging in via proxy, it might always select the wrong one?

If that made sense and you agree, I could look into supporting converting user types to PROXY_USER in the admin user dashboard if you think that is a possible solution (or if you have another way around that or think its not a problem, let me know)?

lufixSch: You are right this could be an issue. I did not think of that.

The ability to migrate users would be a viable solution. 
Another Idea would be to check for an existing user with `PROXY_USER` type before registering a new user. But that would still mean one could never create a user for regular login using the same name as already used by a forward auth user.

Usually when using proxy authentication, forward auth would be the only login method allowed on the service as the only way to reach the page is through the proxy (As soon as automatic login is implemented a user wouldn't be able to reach the login page anyway). Therefore, I don't think of it as a big issue. This doesn't mean it shouldn't be fixed but that the easiest/fastest solution is probably the best one.

IRHM: Sounds good, I think I will look into allowing changing user types in the admin dash. It should allow changing users to any login type which sounds like it could be useful for other login types too, so may be the most helpful solution for everyone.

christaikobo:
I was suggesting to hide/disable logout is because most of the users would completely let forward auth middleware handle authentication once they switch to proxy header authentication. And it is how it's implemented for some of the projects I self host, either logout is hidden or it does nothing other than refreshing the page. 

If there is use case for people to still need the logout button, then by all means we should keep it, it's just that I'm a bit confused:

> With the current login solution Watcharr only checks for the proxy header on login and uses the JWT afterward. So if you want to change users you would still need to press the logout button.

But doesn't that make a loop? For example I click logout and I'm back to the login page where it checks the proxy header and it logs me in again. And the double auth could be confusing to user as well, like if `Jack` logs in authentik/authelia as `Jack`, he is logged in in watcharr as `Jack` as well. He then logs out watcharr and creates a new account `John` in watcharr, but the account is not created in authentik/authelia, then once the current authentication middleware session expires, he can't go back in as `John` because he created the watcharr account but not authentik/authelia account. He will have to log in authentik/authelia as `Jack` first then log out as `Jack` in watcharr and then log in as `John` again every time. If both `Jack` and `John` are already present in authentik/authelia, he should simply logout and log in again as the user he desires in authentik/authelia instead of wathcarr.

I think if user want to logout they should logout the forward auth middleware. IMHO maybe we can redirect user to the forward auth page when they hit the "logout" button in watcharr, they can logout their authentik/authelia session there. How does that sound?

IRHM:
I think you're right @terrytw. When we add the auto login flow (currently requires a button press on login page) then the logout button would indeed be a loop and we should probably hide or alter what it does in that case.

Is there a standard way that auth providers give services (us) a page that we can redirect to, to let the user *actually* logout, or would we just need to add another setting in the admin panel like "Proxy Logout URL" or "Proxy Address"?

I tried with my test authelia setup and if I navigate to the domain its setup at I can indeed click a logout button, if i were to guess, that would be what all of them do.

christaikobo:
@IRHM I have been using authelia as well, so don't know how authentik handles things. AFAIK there is no standard way, services will probably have to add a setting for logout redirection URL. 

Most services don't even bother with that...It's definitely nice to have though.

IRHM:
> @IRHM I have been using authelia as well, so don't know how authentik handles things. AFAIK there is no standard way, services will probably have to add a setting for logout redirection URL.

I suppose we could only show the logout button if that setting is configured then, I think that should be okay.

christaikobo:
For reference, authelia has documents about either a frontend or backend solution for logout. https://github.com/authelia/authelia/commit/100d598a0edfef6980e2cbf2af554f4217d27497 The api document is also available at `/api/` where you host authelia.

But I think the solution we discussed above (which corresponds to the frontend approach in the document) is easy to implement and works just as well.

lufixSch:
Usually forward auth services provide a URL to redirect to, if you want to log out again. Adding this as a feature for logout would be the cleanest solution. Of course the specific structure of that value may vary depending on the use service. This means a setting for the logout URL is needed

As @terrytw said most applications don't bother implementing this feature, but they usually use tokens with a limited lifespan (usually the session). Watcharr instead uses a JWT which doesn't expire. Therefore, a working solution for logout is definitely needed.

christaikobo:
Session expiration can also be handle by authelia/authentik if that is a concern. 

It's ultimately up to you guys, I'm just providing some context.

IRHM:
Thank you @lufixSch and @terrytw, I have merged everything auth related in #736, so closing this one now. If anything doesn't look right please let me know!

lufixSch:
@IRHM Thanks for doing the work making my proof of concept actually usable. I just upgraded my instance with your changes, and it's working very well so far. I'm glad I could help with this.

I'll just take this Opportunity to thank you for creating and maintaining Watcharr. Me and my friends with access to my instance really like it.

IRHM:
> @IRHM Thanks for doing the work making my proof of concept actually usable. I just upgraded my instance with your changes, and it's working very well so far. I'm glad I could help with this.
> 
> I'll just take this Opportunity to thank you for creating and maintaining Watcharr. Me and my friends with access to my instance really like it.

That is awesome, thank you and I am glad you are enjoying the app!

Issue #619, opened 2024-09-04
Feature: people view


AlexPerathoner:
Hello, some more features I'd like to see in the actors/cast view:
- Ability to add certain people to a favorites list
- Ability to filter and sort shows / movies shown. Filter by if they are present in my list; sort by popularity (which I guess is the current way), newest / oldest - Sorting completed in #670

Again I'd be happy to help

IRHM:
Hey @AlexPerathoner, some cool ideas, that would be great!

Do you want to work on both of those ideas?

Also do you have any ideas on how you would want favourited people to show up/be implemented? Would you want them to showup in your watch list or would you prefer they are listed in a popup similar to followed users (we could maybe show followed users and actors/crew in that same popup we already have, differentiating watcharr users and cast with an icon maybe or two separate lists).

For the filters, I'm guessing we can copy the watchlist FilterMenu to the people page and edit it for the filters we want to display there.

p.s. i think showing followed cast in the same menu as other followed watcharr users sounds cool (ofc with some differentiator)?

PR #670, opened 2024-10-23
Sort items in people view

AlexPerathoner:
Relates to  #619
### Changes made
Added sort dropdown to sort items in `/person/[id]` views by Vote count (former method), oldest / newest release date

immagine

### DIFF
--- a/src/routes/(app)/person/[id]/+page.svelte
+++ b/src/routes/(app)/person/[id]/+page.svelte
@@ -170,7 +170,7 @@
           </div>
         </div>
       </div>
-      <div class="filters container">
+      <div class="filters">
         <DropDown
           bind:active={sortOption}
           placeholder="Vote count"
@@ -203,16 +203,14 @@
     align-items: center;
     display: flex;
     justify-content: flex-end;
-    margin-bottom: -55px;
-    margin-top: 35px;
-  }
-  .container {
     margin: 0 auto;
-    min-width: 320px;
     padding-left: 20px;
     padding-right: 20px;
     width: 100%;
+    /* Same as in PosterList */
+    max-width: 1200px;
   }
+
   .content {
     position: relative;
     color: white;
--- a/src/routes/(app)/person/[id]/+page.svelte
+++ b/src/routes/(app)/person/[id]/+page.svelte
@@ -69,13 +69,12 @@
 
   function sortCredits(sortOption: string) {
     if (!credits || !credits.cast) return;
-    let sorted = [...credits.cast];
     switch (sortOption) {
       case "Vote count":
-        sorted.sort((a, b) => b.vote_count - a.vote_count);
+        credits.cast.sort((a, b) => b.vote_count - a.vote_count);
         break;
       case "Newest":
-        sorted.sort((a, b) => {
+        credits.cast.sort((a, b) => {
           const dateA = new Date(a.release_date || a.first_air_date).valueOf();
           const dateB = new Date(b.release_date || b.first_air_date).valueOf();
 
@@ -87,7 +86,7 @@
         });
         break;
       case "Oldest":
-        sorted.sort((a, b) => {
+        credits.cast.sort((a, b) => {
           const dateA = new Date(a.release_date || a.first_air_date).valueOf();
           const dateB = new Date(b.release_date || b.first_air_date).valueOf();
 
@@ -99,7 +98,7 @@
         });
         break;
     }
-    credits = { ...credits, cast: sorted };
+    credits.cast = credits.cast;
   }
 </script>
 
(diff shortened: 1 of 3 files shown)


src/routes/(app)/person/[id]/+page.svelte
```diff
@@ -19,6 +19,9 @@
   let personId: number | undefined;
   let person: TMDBPersonDetails | undefined;
   let pageError: Error | undefined;
+  let sortOption = "Vote count";
+  let cachedCredits: TMDBPersonCombinedCredits | undefined;
+  let sortedCredits: TMDBPersonCombinedCredits | undefined;
```
Comment (IRHM) [inline src/routes/(app)/person/[id]/+page.svelte]: Instead of having two lists, we could just have one `credits` array and sort it in place whenever a the sort is changed?

src/routes/(app)/person/[id]/+page.svelte
```diff
@@ -31,33 +34,64 @@
...
+    try {
+      person = undefined;
+      pageError = undefined;
+      if (!personId) {
+        return;
       }
-    })();
+      person = await getPerson(personId);
+      cachedCredits = await getPersonCredits();
+      sortedCredits = getSortedCredits(cachedCredits, sortOption);
```
Comment (IRHM) [inline src/routes/(app)/person/[id]/+page.svelte]: If we just have the one array of credits, `getSortedCredits` could become `sortCredits`, which would simply sort the one credits array (then also wouldn't have to return anything)

src/routes/(app)/person/[id]/+page.svelte
```diff
@@ -31,33 +34,64 @@
...
-    const credits = (await axios.get(`/content/person/${data.personId}/credits`))
+    let credits = (await axios.get(`/content/person/${data.personId}/credits`))
       .data as TMDBPersonCombinedCredits;
-    credits.cast = credits.cast?.sort((a, b) => b.vote_count - a.vote_count);
     return credits;
   }
+
+  function getSortedCredits(credits: TMDBPersonCombinedCredits, sortOption: string) {
+    if (!credits || !credits.cast) return;
+    let sorted = [...credits.cast];
```
Comment (IRHM) [inline src/routes/(app)/person/[id]/+page.svelte]: with the one array, we wouldn't need to spread the credits into this intermediary array to sort (which I believe can be expensive), we can just run `credits.sort` below (i believe that should work)

src/routes/(app)/person/[id]/+page.svelte
```diff
@@ -31,33 +34,64 @@
...
+    switch (sortOption) {
+      case "Vote count":
+        sorted.sort((a, b) => b.vote_count - a.vote_count);
+        break;
+      case "Newest":
```
Comment (IRHM) [inline src/routes/(app)/person/[id]/+page.svelte:76]: The Newest and Oldest filters seem to be able to break on certain credits lists.

It works on your example page, but for example, can break on [Uma Thurman (/person/139)](http://localhost:5173/person/139).

With her list as an example, if I set the sort to Newest, the list looks right, until I get towards the bottom, where it seems to have just stopped sorting (same thing happens, but opposite with Oldest sort).

It seems to always break after the movie "The Old Guard 2" (in her list specifically), which I don't think has a release date yet, which _might_ be why the sort breaks.

src/routes/(app)/person/[id]/+page.svelte
```diff
@@ -31,33 +34,64 @@
...
+        break;
+      case "Oldest":
+        sorted.sort(
+          (a, b) =>
+            new Date(a.release_date || a.first_air_date).valueOf() -
+            new Date(b.release_date || b.first_air_date).valueOf()
+        );
+        break;
+    }
+    return { ...credits, cast: sorted };
```
Comment (IRHM) [inline src/routes/(app)/person/[id]/+page.svelte]: with the one array, i believe we shouldn't have to return anything - though possibly we may need to reassign credits = credits to get it to update ui (not 100% certain on that)

IRHM (changes requested):
Hi @AlexPerathoner, thank you for taking the time to do this!

I noticed one bug with the sorting and one small thing I think we could improve upon (we could possibly be using just one list instead of two for storing the credits).

Let me know if I can help with anything (or if you want me to finish any of this)!

AlexPerathoner:
Removed the two cached arrays into just one; updated methods to reflect that; updated sort method to treat shows which don't have a release date nor a first air date as unreleased (so they appear first when sorting by Newest and last when by Oldest).

Also noticed that some people have duplicate items in them (because of different `character` and `credits_id` (for example https://api.themoviedb.org/3/person/25072/combined_credits has 59941 twice, once as `Self` and once as `Self - Guest`). Since in the people view there's no reference to the played character yet, I added a filter to remove these duplicates.

IRHM (approved):
This is awesome, thank you!

I'm liking the less duplicates in the person view, it has always annoyed me!

p.s. I made a few tiny changes, didn't find any bugs, just changed around some code/styling.

PR #671, opened 2024-10-23
Filter by items in watch list in people view

AlexPerathoner:
Relates to  #619

### Changes made

#670 should be reviewed first

Added checkbox in people view to filter items to only those which are already in the watch list

| immagine | immagine |
|-------|------|


AlexPerathoner:
Perhaps the filter button on top should be made visible and use that instead?
immagine

IRHM:
Thank you for working on this and for the idea in the first place! I took your code and modified it in #704. It seems more responsive for larger lists now (wrote in more detail in the PR description), I hope you agree if you ever try it out.

I co-authored you on the commit since I used your code as a base, I hope that is okay! I'll close this one in favor of that one.

PR #672, opened 2024-10-23
Color by status

AlexPerathoner:
### Changes made

Added background color on Poster based on watch status

immagine

### DIFF
--- a/src/lib/nav/DetailedMenu.svelte
+++ b/src/lib/nav/DetailedMenu.svelte
@@ -17,7 +17,11 @@
 <Menu
 	conf={{
 		width: "200px",
-		right: "92px",
+		right:
+			page.url?.pathname.startsWith("/person") ||
+			page.url?.pathname.startsWith("/discover")
+				? "12px"
+				: "92px",
 		arrowLeft: page.url?.pathname.startsWith("/search") ? "84px" : "3px",
 	}}
 >
--- a/src/lib/poster/ExtraDetails.svelte
+++ b/src/lib/poster/ExtraDetails.svelte
@@ -25,6 +25,16 @@
 			store.userSettings.ratingSystem === RatingSystem.Thumbs,
 	);
 
+	let statusColorMap = {
+		FINISHED: "var(--status-finished-color)",
+		PLANNED: "var(--status-planned-color)",
+		WATCHING: "var(--status-watching-color)",
+		HOLD: "var(--status-hold-color)",
+		DROPPED: "var(--status-dropped-color)",
+	};
+
+	let backgroundColor = status !== undefined ? statusColorMap[status] : "";
+
 	function formatDate(e: number) {
 		if (!e) {
 			return "Unknown";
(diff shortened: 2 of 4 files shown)


src/lib/poster/ExtraDetails.svelte
```diff
@@ -26,8 +35,8 @@
   }
 </script>
 
-{#if ($page.url?.pathname === "/" || $page.url?.pathname.startsWith("/search/")) && details && dve && dve.length > 0}
```
IRHM: Removing the page check here introduces a bug where:

1. Enable the extra details
2. Go to another users watched list
3. Can see extra details on their page, but can't view <DetailedMenu> to toggle this off/on

I think we could add it back in?

src/lib/nav/DetailedMenu.svelte
```diff
@@ -17,7 +17,12 @@
   $: dve = $wlDetailedView;
 </script>
 
-<div class={`menu${$page.url?.pathname.startsWith("/search/") ? " on-search-page" : ""}`}>
+<div
+  class={`menu${$page.url?.pathname.startsWith("/search/") ? " on-search-page" : ""}`}
+  style="right: {$page.url?.pathname === '/' || $page.url?.pathname.startsWith('/search/')
```
IRHM: Was this to fix an issue? (I couldn't find if there was an issue with old code to review)

IRHM (changes requested):
Hi @AlexPerathoner thank you for implementing this!

I had some comments on some of the code. One of them I was unsure of if I was testing correctly, if it fixed anything im not seeing?

When I find some more time, I wanted to look into the colors too. I may just tweak around with them to see if we can get them to hide less of the poster while still being legible (like the old color). If not, I may implement a setting to allow toggling this on/off.

src/lib/poster/ExtraDetails.svelte
AlexPerathoner: I see. I had added it in to make the extra details appear on all lists of movies where I could be interested in the color by status. That's also why on the other hand I then added the line above.
I'll search for a different way to do it

IRHM: Sounds good @AlexPerathoner, do you think we should just support showing ExtraDetails for posters when viewing other users lists? In that case we could just show the ExtraDetails button in nav on that page too, but I'm not sure if that could get confusing for users (if it's showing, it may seem like it's because it is on their list too), what do you think?

AlexPerathoner: I personally don't think it would be confusing. After all it's just some extra details on items, not necessarily from their own list. Also with the color they can immediately see if it is indeed on their list

IRHM: @AlexPerathoner That sounds good to me

src/lib/nav/DetailedMenu.svelte
AlexPerathoner: When enabling the detailedMenu item on the top, if there are not enough items on the right, without that check it appears like this:
immagine

src/lib/poster/ExtraDetails.svelte
AlexPerathoner: I had misunderstood what meant here. It's indeed quite confusing, I've removed it from the other people's lists

AlexPerathoner:
Any update on this?

PR #704, opened 2024-12-01
Person: 'On my list' filter support

IRHM:
### Changes made

Port for 'On my list' checkbox for person page based on #671 with changes.

@AlexPerathoner I moved some stuff around because I was noticing a jitter when untoggling the checkbox on certain pages (actors with long list of credits). I tried a few solution and ended up realising it was probably just something that would be solved when I get time to look at implementing a virtual list wherever we display <PosterList>s.

For now, I changed the code to hide posters with css to avoid the {#each} block triggering again, which seems to have made it faster. I hope you agree with this solution.


AlexPerathoner:
Amazing, faster is always great!

PR #711, opened 2024-12-07
fix: time parsing issue of content video response

0nlyJ2W:
This is a fix for the time parsing issue when requesting a detailed view of movies and tv shows.
The origin of this issue was probably a change in the tvdb api response of the video content.
The published at string was formatted like this:
```json
"videos" : {
    "results" : [ {
      "published_at" : "2021-02-25 11:45:24 UTC",
    }]
}
```
Which caused a failure parsing the response as type struct due to a different expected time format.

The solution was to create a custom time struct and override the UnmarshalJSON interface to support the given format.

### DIFF
--- a/server/tmdb.go
+++ b/server/tmdb.go
@@ -7,6 +7,7 @@ import (
 	"log/slog"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


0nlyJ2W:
They have reverted the time format. This fix is obsolete now.

IRHM:
Glad it is fixed, thanks for taking the time regardless to get a fix out so quickly!

PR #736, opened 2024-12-28
Trusted header authentication

IRHM:
### Changes made

Work added on top of #632 (used as base)

Main changes:

- Added more related settings, now presented in a modal:
  image
- Configurable Auto Login (simply attempts a trusted header login when anyone reaches login page, overrideable by adding `noAuto=1` query parameter to login page url.
- New logout modal for proxy users (SSO logout button redirects to the Logout URL setting):
  image


IRHM:
@lufixSch I moved some of the code around to new files and renamed some of your original code, so made a new PR instead of overwriting yours, I hope that is okay.

PR #819, opened 2025-02-19
feat: added nav menu item close on click outside

### LINKED ISSUE #813: Close nav menus after clicking into page
**Idea Overview**
With a nav menu open (detailed view, sort, filter, etc) clicking anywhere on the page should close the open menu.

**Is your feature request related to a problem? Please describe.**
It's annoying when it stays open.

**Describe the solution you'd like**
Click anywhere on the page to close any open nav menus.

IMPORTANT: This feature should not affect clicking links/buttons on the page with a nav menu open.

**Additional context**
Reported by chiptune on matrix chat.

antoniosarro:
Navbar menus now close when anywhere outside the menu is clicked.

Related to issue #813


src/lib/util/helpers.ts
```diff
@@ -409,3 +409,49 @@ export function mouseOverEl(el?: HTMLElement) {
...
 }
+
+/**
+ * Helper function to execute a callback when the user clicks outside of a specified element.
+ *
+ * @param {HTMLElement} node - The element within which click events are ignored.
+ * @param {() => void} callback - The function to be called when a click occurs outside the specified element.
+ * @returns {{ destroy: () => void }} An object with a `destroy` method to remove event listeners and stop listening for clicks outside the element.
+ */
+export function clickOutside(
```
IRHM: Actions currently each have their own file under the `src/lib/actions` folder. I think we could move this to it's own file there to stay organized with where actions are.

src/routes/(app)/+layout.svelte
```diff
@@ -222,7 +222,10 @@
 					{/if}
 				</button>
 				{#if detailedMenuShown}
-					<DetailedMenu />
+					<DetailedMenu
+						clickOutsideCallback={() =>
```
IRHM, line 227: There appears to be a bug introduced with each Menu that uses this callback. You can no longer press the menu button used to open the menu to close it.

It seems that when you try to close a menu with its menu button in the nav (using the sort menu as an example, happens to all):
1. The `clickOutsideCallback` will set `sortMenuShown` to `false`
2. The menu button `onclick` will set `sortMenuShown` back to `true`.
3. Menu stays open and cannot be closed via the initial button used to open it.

src/lib/util/helpers.ts
```diff
@@ -409,3 +409,49 @@ export function mouseOverEl(el?: HTMLElement) {
 	}
 	return el.matches(":hover");
 }
+
+/**
+ * Helper function to execute a callback when the user clicks outside of a specified element.
+ *
+ * @param {HTMLElement} node - The element within which click events are ignored.
+ * @param {() => void} callback - The function to be called when a click occurs outside the specified element.
```
IRHM: Just a small comment styling thing, I think the dashes (`-`) should be removed from the comments for `node` and `callback` parameters (makes it show as a bullet point when markdown rendered).

src/lib/Menu.svelte
```diff
@@ -15,12 +16,14 @@
 	interface Props {
 		children: import("svelte").Snippet;
 		conf?: MenuConfig;
+		clickOutsideCallback?: () => void;
 	}
 
-	let { children, conf }: Props = $props();
+	let { children, conf, clickOutsideCallback = () => {} }: Props = $props();
```
IRHM: `clickOutsideCallback` shouldn't be given a default value (empty anonymous function).

Instead, we can leave it as undefined if it isn't set, then inside the `clickOutside` action we can check if `callback` is of type `function`. If not, then we can disable the action as to not add unnecessary event listeners, etc.

IRHM (changes requested):
Hi @antoniosarro, thank you for taking the time to work on this feature.

I found one bug and a few other things that I think would improve upon the code. Let me know if there's anything I can help with further or if I am misunderstanding anything in my comments.

Thanks!

src/lib/util/helpers.ts
antoniosarro: Yes, that's definitely a better place to put it

src/routes/(app)/+layout.svelte
antoniosarro: I think that the fastest and simplest method to solve this bug is to add a class to the button that opens the menu, and exclude clicks on that particular class within the action. Tell me what you think and I'll take a look

IRHM: > I think that the fastest and simplest method to solve this bug is to add a class to the button that opens the menu, and exclude clicks on that particular class within the action. Tell me what you think and I'll take a look

It seems like that may be the only solution here. I think it would force us to use different ids for each button and pass them down to each menu individually as to not break clicking on a different nav menu button to switch over to it. Do you think it could work like that (if necessary)?

We should probably add unique `id`s to the buttons in a certain format (like: id="menuButtonNavSort").

I think we'd also have to check if the click target was an element inside of the menu button (eg: the icon; an svgs `<path>`)

antoniosarro:
I made the changes and it seems to work as it should, I don't know if it is necessary to define the id used to exclude buttons as a const somewhere so that it is easier to modify it in case for each occurrence.
I also made the other changes you listed

IRHM:
Hi @antoniosarro, thanks for making them changes. Looks like I am seeing a new bug:

Clicking into the nav or body just below the nav still closes to popup, but for some reason clicking anywhere in the posters wrapper or on posters doesn't close the popup (eg: space around posters, poster rating/status button). Same on other pages too, just the top part of the page will close them, nothing in the actual body.

antoniosarro:
Yes, I saw that it was a bug that I had missed because when I tested it the poster wrapper was not present.
Now I checked and it doesn't give that problem anymore.
I checked that repeatedly clicking on the icon, clicking within the open menu and clicking within the page works and they all seem to work as they should.

PR #822, opened 2025-02-22
feat(theme): add system preference handling to theme configuration

### LINKED ISSUE #818: Sync Interface Theme with System Color Preference
**Idea Overview**
Enable system color preference tracking: allow the interface to respect and adapt to the user's system-wide dark mode setting.

**Is your feature request related to a problem? Please describe.**
Currently, the interface only provides a toggle between dark and light modes, which may not be in sync with the user's system preferences. This can lead to inconsistencies in the user experience, particularly for users who rely on their system settings to control their visual environment.

**Describe the solution you'd like**
Add an option to track the system color preference and automatically adapt the interface theme accordingly. This would ensure that the interface is in sync with the user's preferred visual settings.

antoniosarro:
- Modified the store to handle system `theme` changes and manage deletion and rehydration of the store.
- Added a button to select the `system` theme in the settings.
- Updated the `toggleTheme` function to listen for changes in the `prefers-color-scheme` preference when the 'system' theme is selected.

Fixed #818

### DIFF
--- a/src/lib/util/helpers.ts
+++ b/src/lib/util/helpers.ts
@@ -1,3 +1,4 @@
+import { browser } from "$app/environment";
 import { store } from "@/store.svelte";
 import {
 	UserPermission,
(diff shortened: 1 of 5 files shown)


IRHM (approved):
Great work, thank you for getting this added in, it works great!

I made a few changes that I think improve the code a little:

- Moved `system` theme button to top/first (since its default).
- Moved theme logic to new `theme.ts` file & renamed some variables.
- Store: Revert change that sets updateStore=true when rehydrating theme (not necessary since we update store manually)
- toggleTheme: Removed `return` at the top when the media query isn't defined (probably will never occur, but incase it does for a user, the rest of the code should still be able to run for setting other themes).
- toggleTheme: Now always removes the eventListener before adding it to ensure there are no cases where we add multiple event listeners.

PR #862, opened 2025-04-07
Support for importing ryot 8 files

IvanBeke:
### Changes made

Ryot 8 has made some changes to its json export file and the current import process is out-of-date. To fix the import process I've made the following changes:
- `media` renamed to `metadata`
- Shows can be given an overall rating


IRHM (approved):
This is great work, thank you for fixing this! Sorry it took so long to get this merged.

PR #880, opened 2025-04-29
feat: improve imdb import

jigglycrumb:
### Changes made

This PR makes the IMDb importer more robust by identifying more media types like Direct-to-Video movies, TV Mini Series and more which are all not regconized and thus skipped by the current importer.

It also makes a tiny UI change by adding the year of the imported movie to the "Multiple Results Found" dialog, which makes identifying the actually imported movie more easy and should help to reduce mismatches.

Here is an example screenshot - the new part is the year display in parentheses after the "Select multiple items for ..." - the year and parentheses are only shown when the data is available.

2025-04-28-142835_hyprshot

### DIFF
--- a/src/routes/(app)/import/process/+page.svelte
+++ b/src/routes/(app)/import/process/+page.svelte
@@ -1,4 +1,4 @@
-<!-- 
+<!--
   /import/process is for processing the
   selected files data. Here it will be
   displayed and imported.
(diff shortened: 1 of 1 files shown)


IRHM (approved):
Awesome work, thanks for doing this!

(sorry it took so long to merge, I've been busy working on a new update which has been roadblock after roadblock sadly)

PR #913, opened 2025-07-03
Bug fix on log message

ParksideParade:
Small bug fix so that Twitch init error is logged only when err is not empty


IRHM:
Hi @ParksideParade, thanks for opening this PR. It looks like a second commit might have been accidentally pushed up here?

PR #921, opened 2025-07-15
Fix Twitch init error logging bug

ParksideParade:
### Changes made

One line fix to correct the error log debug message.


IRHM (approved):
Thanks for this!

PR #922, opened 2025-07-15
fix bug on return from blurhash.Encode

ParksideParade:
### Changes made

One-line fix when calling blurhash.Encode, as (unlike Decode) the second return is the error.

https://github.com/bbrks/go-blurhash/blob/master/encode_test.go#L34


IRHM (approved):
Another great catch, thank you!

PR #969, opened 2025-09-19
Update docker-compose.yml

BoGu5:
port number of docker container is 3000. Docker compose file now matches this.


IRHM:
Hi @BoGu5, thanks for taking the time to fix this.

The current compose file already has the correct port, you may be getting confused with the UI server port which is only for internal use inside the container. The current port (3080) is what should be used.

PR #1035, opened 2026-03-12
feat: initial support for books

### LINKED ISSUE #382: Tracking of book reading
Book tracking, i.e. tracking progress of books/read/unread books etc. would be really nice to have. Would you consider adding book reading/tracking as a feature?

Watcharr tracks everything else, books are the only thing missing :)

Bnyro:
### Changes made
- this adds support for books via OpenLibrary
- the only UI part of this that is already working is the search (UI icon still buggy though, will fix that later)
- the backend part should be close to done (I still have to create the router for book details and author details, but the OpenLibrary API client I implemented already supports it, so it should be easy enough)

## Notes
I'm already opening this to share my progress and hopefully get some feedback if this is going into the right direction or if there's something fundamentally wrong with my logic?

Anyways this took much longer than I expected (already worked more than 6 hours on this), so taking some break now. Not sure when I find time to continue working on this, probably some time next week.

(one day) closes #382

PS: I noticed that the [development guide](https://github.com/sbondCo/Watcharr/blob/dev/doc/docs/installation/for-development.md) doesn't work, as there's no `server` command in `package.json`. I've used the steps from the installation instructions now for testing it which worked fine.

PPS: I'll remove the `when` library dependency again probably, it doesn't seem like it's able to properly parse the date format from `OpenLibrary` anyways.

(Note to self: OpenLibrary docs are at https://openlibrary.org/developers/api)


Bnyro:
Okay, we're getting closer. Below a list of what still doesn't work:
- ~Ratings don't work yet, e.g. giving a thumbs up to a book does nothing yet (although in theory, the required code is already there, but there's still something broken)~
- ~The date parsing library (`when`) should be replaced by something that actually works~

Open questions (what's your opinion about that @IRHM?):
- Maybe add an author search route? Currently, I only added a book searching route, I'm not sure if author search is worth it.
- Biographies returned by OpenLibrary are in Markdown format, but we currently show it as plain text instead of rendering it as HTML. To support rendering properly, we would need to add a dependency to something like https://github.com/gomarkdown/markdown.

IRHM:
>     * Maybe add an author search route? Currently, I only added a book searching route, I'm not sure if author search is worth it.

Book searching is probably good enough for now, if you want to add author searching feel free, though we may need to think about how it will be displayed in the UI since we already have a 'People' filter.

>     * Biographies returned by OpenLibrary are in Markdown format, but we currently show it as plain text instead of rendering it as HTML. To support rendering properly, we would need to add a dependency to something like https://github.com/gomarkdown/markdown.

I didn't expect they'd use markdown, I guess we should add parsing to html.

I don't mind if you do this in this PR or if we leave it for a later date, since it looks like most authors don't have any formatting for their bio (?), and it's no big deal since we can always add it in in the future.

If you do decide to add it in, we should probably add sanitization, like how gomarkdown describes: https://github.com/gomarkdown/markdown?tab=readme-ov-file#sanitize-untrusted-content

I haven't looked into it too much, but this library may be better (im basing that off stars and because hugo uses it): https://github.com/yuin/goldmark/

Bnyro:
> Book searching is probably good enough for now, if you want to add author searching feel free,
>
Yeah, I don't think it's worth it either, so I'm fine with not adding it. Actually, if searching e.g. "Dan Brown" in the books section, it already returns books by that author, so there's really no need for an additional author search.

> since it looks like most authors don't have any formatting for their bio (?)
>
I think the book storylines use Markdown, and not the biographies that. Probably mixed that up, sorry. About half of them used Markdown iirc.

> If you do decide to add it in, we should probably add sanitization, like how gomarkdown describes: 
>
Yeah, we should definitely do that. I'll look into the Markdown rendering the coming days (probably tomorrow)

Bnyro:
Hey @IRHM, I just rebased the PR.

I would be very happy about a review if you get the chance :)

IRHM:
Double thanks @Bnyro 

I'll try to have a good look this weekend!

server/domain/media.go
```diff
@@ -20,6 +22,9 @@ const (
 	MediaTypeTMDBPerson MediaType = "tmdb_person"
 
 	MediaTypeIGDBGame MediaType = "igdb_game"
+
+	MediaTypeGenericBook       MediaType = "generic_book"
+	MediaTypeGenericBookAuthor MediaType = "generic_book_author"
```
IRHM: Should the `generic` prefix be changed to `openlibrary`?

In Medias `GetId()` you return OLID for MediaTypeGenericBook.

(same for frontend of course)

server/feature/book/entity/author.go
```diff
@@ -0,0 +1,42 @@
+package entity
```
IRHM: This `Author` struct may be better inside our `domain` package (since it isn't directly database related).

server/media/openlibrary/openlibrary.go
```diff
@@ -0,0 +1,347 @@
...
+const resultsPerPage int = 20
+
+func NewOpenLibrary() OpenLibrary {
+	return OpenLibrary{}
+}
+
+// Make an API request at the provided `path` and the given params `p`.
+//
+// If the method returns no error, it is guaranteed that `resp` now contains the JSON response.
+func (o *OpenLibrary) APIRequest(path string, p map[string]string, resp interface{}) error {
```
IRHM: Ideally this function is local only.

src/lib/poster/PersonPoster.svelte
```diff
@@ -1,33 +1,37 @@
...
 		? `https://image.tmdb.org/t/p/w300_and_h450_bestv2${path}`
-		: undefined;
-	const link: `/person/${number}` | undefined = id
+		: path
+	: undefined;
+const link = id
+	? mediaType === MediaTypeE.tmdbPerson
 		? `/person/${id}`
-		: undefined;
+		: `/book/author/${id}`
```
IRHM: Instead of having `/book/author` be the fallback, can we add a check of its mediaType for openLibraryAuthor?

You could put the whole statement into a $derived.by() so you can use if guard statements or a switch so its more readable.

src/lib/poster/Poster.svelte
```diff
@@ -136,6 +140,15 @@
 			} else {
 				return `https://image.tmdb.org/t/p/w500${media.extPosterPath}`;
 			}
+		} else if (media.type == MediaTypeE.genericBook) {
+			if (watched) {
```
IRHM: You shouldn't need this `if (watched)` check because it should return a poster in the first if statement of this derived if we have one (tmdb stuff hasn't been migrated to use the Images table so it is still doing it the old fashioned way). 

It can be replaced with the same logic as for games below (just returning an external image url).

server/database/entity/book.go
```diff
@@ -0,0 +1,39 @@
...
+	OLID string `json:"olid" gorm:"uniqueIndex"`
+
+	// List of edition ISBNs, separated by "|"
+	ISBN          string  `json:"isbn"`
+	Title         string  `json:"title"`
+	Storyline     string  `json:"storyline"`
+	RatingAverage float64 `json:"ratingAverage"`
+	RatingCount   int     `json:"ratingCount"`
+	// although this could be derived by the ID for OpenLibrary, this is stored to allow adding more book search providers in the future where URLs can't be derived from the ID
+	CoverUrl string `json:"coverUrl"`
```
IRHM: I think we should remove this and create the full external cover url in the frontend (eg: how it is done for games). Since we can derive the image from the OLID, we can just use that as you say, but directly in the frontend.

If it is needed in the future we can add it back in on that date for only the services that require it similar to how our `Games` table has a `cover_id`, not the full url since most services allow adding image width/height, etc to the url (which can be useful for the frontend to decide when it constructs the full final url).

IRHM (changes requested):
@Bnyro Thanks for the hard work and effort put into this PR, it is great!

I added a few changes that I think make sense.

Also apologies, the `dev` branch has some more changes causing merge conflict. I was going to rebase for you, but didn't want to remake your commits and break your local branch.

server/domain/media.go
Bnyro: Yeah, I initially wanted this to be compatible with other book providers too, but I think it would be better to add a different media type for other source (e.g. comic libraries)

src/lib/poster/PersonPoster.svelte
Bnyro: > Instead of having /book/author be the fallback, can we add a check of its mediaType for openLibraryAuthor?
>
I thought so, that probably broke during rebasing. The `$derived.by` approach is a very good idea, looks much cleaner now. Thanks :+1:

src/lib/poster/Poster.svelte
Bnyro: Makes sense, thanks for the info :+1:

Bnyro:
> @Bnyro Thanks for the hard work and effort put into this PR, it is great!
> 
> I added a few changes that I think make sense.
> 
> Also apologies, the `dev` branch has some more changes causing merge conflict. I was going to rebase for you, but didn't want to remake your commits and break your local branch.
>
Thank you for the review!

Rebased, applied your suggestions and fixed some small bugs I found on the way.

I've squashed everything into two commits, one for the backend and one for the frontend.

I tested this a bit, but unfortunately I managed to get my residential IP blocked because I tried to parallelize loading book details, so I couldn't do a final test :/ I'll probably get a new IP from the ISP tomorrow automatically, will test it again then.

Bnyro:
Okay, I got unblocked from OpenLibrary.

Everything seems to work now. The only flaw I found is that the "credits" section in the author page (i.e. books by that author) OpenLibrary often returns some random editions of the book that don't have any cover. But usually there are enough with a cover to make it look fine.

PR #1055, opened 2026-06-06
feat: show spoilers for finished episodes

goestav:
### Changes made

- always display spoilers for finished episodes
- ~~only display the rating button when the spoiler overlay is removed~~
  allow users to rate episodes regardless of spoiler state or episode status[^1]
- let users change the episode status when the spoiler overlay is visible

[^1]: I'm not sure what the expected behavior here is supposed to be, I changed this so it behaves similar to how movies currently work.


IRHM:
Hi @goestav thank you for opening this PR!

I'm going to close this one and merge https://github.com/sbondCo/Watcharr/pull/1066 which has (what I think are) a few improvements (i've added you as Co-Author and will tag you in the changelog for getting it working).

Thanks!

Issue #1048, opened 2026-04-25
Transition from FINISHED to WATCHING when new episodes are found


WhyNotHugo:
**Idea Overview**

When a show has new episodes, auto-transition it from FINISHED -> WATCHING.

**Is your feature request related to a problem? Please describe.**

After I've marked a show as finished, if new episodes come out, there's no way for me to know that a new episode has come out. Two factors are involved here:

1. No mechanism fetches updates for shows: if I don't explicitly navigate to its page, it won't even show a new episode. But I'd only navigate to its page if I knew there are new episodes in the first place. For this aspect, see: https://github.com/sbondCo/Watcharr/issues/1049

2. If I navigate to the page and new data is fetched, the show remains marked as "finished", despite new episodes being available which haven't been watched.

**Edit**

After I rewrote this issue multiple times, I realised that it was two distinct feature requests, so split half of it into https://github.com/sbondCo/Watcharr/issues/1049

binnichtaktiv:
+1 would be nice if it happens automatically

Issue #1008, opened 2026-01-02
Chronological view of upcoming episodes / movies + unwatched content


wolfgang101:
**Idea Overview**
Show a chronological list of upcoming episodes / movies and unwatched content.

**Is your feature request related to a problem? Please describe.**
Not related to a problem, but found this a really nice feature in other trackers. Makes it really easy to keep track of what's on the "to watch" list and what will be available soon.

**Describe the solution you'd like**
Could be a standalone page or integrated as section into the overview or discover page.
- Could show what's airing within the next e.g. 14 days (or e.g. the next X items, or everything that has a known future air date)
- I like the Simkl implementation that not only shows future episodes / movies but also keeps already aired content that is not marked as watched. IMO one of the best features in a tracker, as it makes it much easier to keep track and removes mental clutter, as I don't have to worry about missing anything.

**Additional context**
One service that implements this really well is Simkl - as a reference.

fjacques-2025:
## Proposal: an "Up Next" section on the overview page
A horizontal row (reusing the existing `HorizontalList`) at the top of the overview,
showing — for each in-progress show — its **next unwatched episode**: still/poster,
show title, SxxExx, episode name. Each card has quick actions:
- ✓ **Mark watched** → marks that episode and the card advances to the next unwatched
  episode (or drops off the row once the show is caught up).
- ⭐ **Rate** → rates that episode (same episode endpoint).

This would also naturally cover #903 (quick-mark the next episode without opening the
show and scrolling to the season).

## Suggested scope split
- This proposal is only the **unwatched "to watch"** part.
- The **chronological "airing soon"** part from the original request needs fresh air
  dates, which depends on a periodic content-refresh mechanism that doesn't exist yet
  (see #1048 / #1049). I'd suggest tracking that half separately so "Up Next" isn't blocked.

## A few questions before I start
1. **Placement**: a section on the **overview** page (my preference), a **standalone page**,
   or on **discover**?
2. **Computing "next episode"** across many shows needs per-season episode structure.
   Would you prefer a small **backend endpoint** (e.g. `GET /watched/upnext`) that returns
   the next unwatched episode per in-progress show in one call — or a **frontend-only**
   approach reusing the existing content/season endpoints?
3. **Definition of "next"**: the first gap in watched episodes, or simply the episode
   right after the latest watched one?

Happy to implement this if the direction sounds good — could it be assigned to me?
I'd start on the frontend once placement + the endpoint question are settled.

[...]

Issue #903, opened 2025-06-13
Quickly mark Episode as watched


iLuckyW:
**Idea Overview**
It would be pretty useful to have a quick access button to mark the next episode of a show as watched. Currently, it involves a lot of clicks and scrolling and thus takes a bit of time.

**Describe the solution you'd like**
I would prefer to either avoid scrolling down after clicking on a show banner to find the next episode, or have a little button on the home screen show banner itself to mark the next episode as finished.

**Describe alternatives you've considered**
It would also be possible to make the UI elements in movable. Currently, I need to scroll past the Cast, Similar and Activity section to mark an episode as finished.

**Additional context**
I don't know if you are familiar with the Ryot interface, but there is a similar quick action to mark an episode as watched.

ph-256:
I was just about to suggest that as well, perhaps to make it easier for implementation since it already tracks the last watched episode, perhaps it might just show on the main page UI show card the most current unwatched episode with a checkmark next to it to mark it as watched and then update the show card with the new most current unwatched episode, or even better would be a dropdown with such layout, so you can quickly select the already finished ones
Episode 1 [✔]
Episode 2 [✔]
Episode 3 [     ]

dlong500:
IMO marking movies/episodes watched is the single most important function of an app like this, so I absolutely agree that a single button to mark as watched is a very important feature.  Second to that would be a customizable interface so people could choose which functions are more easily accessible.  But really marking something as watched should only ever be one click away.  Any type of dropdown to get to an action is always going to involve at least two clicks and adds friction.

PR #1069, opened 2026-07-20
feat: "Up Next" row on the overview (next unwatched episode per show)

fjacques-2025:
## What
Adds an "Up Next" row at the top of the overview: for each in-progress (Watching)
show, it shows the next unwatched episode with a quick "✓ Watched" action that
marks it and advances the card to the following episode.

Relates to #1008 (the unwatched "to watch" part) and #903 (quick-mark next episode).

## How
- **Backend**: new `GET /watched/upnext` returning the next unwatched episode per
  Watching show (`server/feature/watched/watched_upnext.go`); scans seasons from
  the highest watched one using cached TMDB season details. `ContentProvider` gains
  `SeasonDetails`.
- **Frontend**: `UpNext.svelte` (a `HorizontalList`) on the overview; "✓ Watched"
  reuses `POST /watched/episode` then refreshes.

## Design note
Episodes are offered regardless of TMDB `air_date`: that date is the TV broadcast
schedule and doesn't reflect streaming/replay availability (a fully streamable show
can still have "future" broadcast dates). The broadcast date is shown on each card,
flagging not-yet-aired episodes — broadcast viewers know to wait, streaming viewers
can mark ahead.

## Draft — open questions
- Placement: overview row, standalone page, or discover?
- Is a dedicated `/watched/upnext` endpoint preferred over computing client-side?
- Definition of "next": currently first unwatched episode from the latest watched season.

Not included yet: rating from the card; the "upcoming calendar" half of #1008
(depends on a periodic refresh mechanism, #1048/#1049).

## Testing
Ran locally (Go + SvelteKit). Marking advances the card; the row clears when a show
is fully watched.

### DIFF
--- a/server/feature/watched/router.go
+++ b/server/feature/watched/router.go
@@ -36,6 +36,7 @@ func (r *Router) AddRoutes() {
 	watched := r.br.Router.Group("/watched").Use(authmiddleware.AuthRequired(nil, r.br.Cfg))
 
 	watched.GET("", router.PaginatedRequest(false), r.GetWatchedList)
+	watched.GET("/upnext", r.GetUpNext)
 	watched.GET(":id/:username", router.PaginatedRequest(true), r.GetPublicWatchedList)
 	watched.POST("", r.AddWatched)
 	watched.PUT(":id", r.UpdateWatched)
@@ -43,6 +44,17 @@ func (r *Router) AddRoutes() {
 	// TODO Move add/delete watched from tag to the `tag` package (the service code is there so the route may as well be under there, also avoids a circular dep).
 	watched.POST(":id/tag/:tagId", r.AddWatchedToTag)
 	watched.DELETE(":id/tag/:tagId", r.DeleteWatchedFromTag)
+}
+
+// Get the "Up Next" list: next unwatched aired episode per in-progress show.
+func (r *Router) GetUpNext(c *gin.Context) {
+	userId := c.MustGet("userId").(uint)
+	items, err := r.s.UpNext(userId)
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, router.ErrorResponse{Error: "failed to get up next list"})
+		return
+	}
+	c.JSON(http.StatusOK, items)
 }
 
 // Get our (logged in user) watched list.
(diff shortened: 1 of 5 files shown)


IRHM:
Hi, thank you for taking the time on this.

I'll have to think about implementation, I have a feeling this should wait until after we have refreshing of our own data to avoid having to rewrite this feature to use it later.

fjacques-2025:
Thanks, that makes sense — I'd rather it fit the eventual data-refresh work than be
rewritten later.

For what it's worth, the current implementation is fairly decoupled: it reads the
existing watched episodes and only pulls season details on demand via the existing
content service/cache. So once self-refreshed/stored episode data lands, swapping the
"next episode" lookup to read from it should be a localized change rather than a
rewrite — happy to rework it that way when the time comes.

I'll leave this as a draft for reference. Is the data-refresh mechanism (#1048/#1049)
on the near-term roadmap, and would help on it be welcome? I'd gladly look into it if
it unblocks this.

fjacques-2025:
Update — I've been running this on my own instance in daily use, and it's working
well. Current state of the feature:

**"Up Next" row on the overview** — for each in-progress show, its next unwatched
episode, with:
- **✓ Watched** — marks it and advances the card to the next episode (also covers #903).
- **Drop** — one click sets the show to Dropped so it leaves the row, without opening it.
- **Season episode count** — e.g. `S1.E7/30`.
- **TV air date** on each card. Episodes are *not* filtered on TMDB `air_date` (that's the
  broadcast schedule, not streaming availability), so it works for streaming/replay while
  still flagging not-yet-aired episodes.

Capture d’écran 2026-07-25 à 09 04 05

Backend is a single `GET /watched/upnext`; no schema changes.

Still a draft — I understand you'd rather land this after the data-refresh work so it
doesn't need reworking, happy to align it with that when the time comes. Sharing the
finished state (and that it's been solid in daily use) in case it's useful.

fjacques-2025:
Small addition (pushed): the Up Next row now follows the **same sort as the watched list** (`GetUpNext` binds `sort`/`sortDir` and applies the shared `watchedRefineSort` scope; the frontend passes the current sort and re-fetches on change). Still in daily use on my instance.

fjacques-2025:
Kept this current — rebased onto v4.2.x and migrated to the new `req` helper, still in daily use on my instance. I know it's gated on the data-refresh work; just flagging it's maintained and ready whenever. Happy to push the rebased branch. 🙂

PR #1078, opened 2026-07-25
fix(search): don't trim the query while typing (drops trailing space)

fjacques-2025:
## Bug
Typing a multi-word query with a brief pause after a space drops the space:
e.g. typing `Mon voisin` (pausing after the space) becomes `Monvoisin` and returns
no results. It's intermittent — it depends on the search debounce firing while the
trailing space is the last character typed.

## Cause
In `src/routes/(app)/+layout.svelte`, the debounced search handler trims the input
value (`target.value.trim()`) before building the `/search` URL. The search input is
bound to `store.searchQuery`, which is reset from the URL query after each navigation.
So when the debounce fires right after a trailing space, the trimmed value (`Mon`) is
written back into the input, dropping the space the user just typed; the next keystroke
then concatenates (`Monv…`).

## Fix
Keep the value exactly as typed for the query, and only skip whitespace-only searches
(`if (!query.trim()) return;`). Nothing trimmed flows back into the bound input.

## Testing
Typing `Mon voisin` with a pause after the space now keeps the space and returns results.

### DIFF
--- a/src/routes/(app)/+layout.svelte
+++ b/src/routes/(app)/+layout.svelte
@@ -68,8 +68,13 @@
 		searchTimeout = window.setTimeout(
 			() => {
 				const target = ev.target as HTMLInputElement;
-				const query = target?.value.trim();
-				if (!query) return;
+				// Keep the value exactly as typed (do NOT trim here). The input is
+				// bound to store.searchQuery, which is reset from the URL query after
+				// each debounced navigation; trimming would drop a trailing space the
+				// user just typed, making "Mon voisin" collapse into "Monvoisin".
+				// We still skip whitespace-only queries below.
+				const query = target?.value ?? "";
+				if (!query.trim()) return;
 				const currentSearchType = page.url.searchParams.get("type");
 				const searchParams = new URLSearchParams({
 					query: encodeURIComponent(query),


fjacques-2025:
Just confirmed this still reproduces on v4.2.x and the fix still applies cleanly. Any chance of a review when you get a moment? Happy to rebase if needed. 🙂

lewisp95:
can you please implement this it's a very annoying issue especially for me as I use an onscreen keyboard due to physical disability.

fjacques-2025:
@lewisp95 thank you for speaking up — and sorry you've been putting up with this.

The onscreen-keyboard angle is worth spelling out, because it changes what the bug *is*. The search navigates on a debounce (400 ms, 800 ms on touch), and the space is lost only when that timer happens to fire while a trailing space is the last character typed. For most people that's occasional, which is why it reads as a papercut. If every keystroke is several hundred milliseconds apart, it isn't occasional — it's every space, every time, and the query you get back is a word you never typed. That's an accessibility problem rather than an annoyance.

I've just re-checked against the current `dev` (v4.2.2): `const query = target?.value.trim();` is still there, so it still reproduces, and the patch still applies cleanly — the file hasn't been touched since the lint pass on 29 July, and not on these lines.

@IRHM — no pressure on my bigger PRs, I know they need design decisions. This one doesn't: 1 file, +7/−2, five of those lines being a comment. The typed value is kept as-is only for building the URL, and whitespace-only queries are still skipped, so nothing else changes. Would you be willing to take a look when you get a moment? Happy to rebase it, reword it, or fix it differently if you'd rather. 🙂

Issue #801, opened 2025-02-12
Multi-language support


IRHM:
Multi-language support.

Not sure how this will all pan out, but an overview of what I currently think we need to get done:

- [ ] Setup Weblate instance
- [ ] Weblate automations (getting translations into this repo in useable format)
- [ ] Integrate i18next for using translations in frontend app (involves moving all current strings into an en-us translation file and using them from there and as default fallback)
- [ ] #442 (this might require structural changes depending on the implementation for caching)

### Discussed in https://github.com/sbondCo/Watcharr/discussions/800

<div type='discussions-op-text'>

<sup>Originally posted by **kiwinh0** February 11, 2025</sup>
When will we have multilanguage?
I would be willing to collaborate with several languages</div>

deadLocks21:
Hey, do you think I can start working on the i18next implementation, before you create an instance of Weblate or something?

IRHM:
> Hey, do you think I can start working on the i18next implementation, before you create an instance of Weblate or something?

Hi @deadLocks21, I believe you can and that would be a big help.

I think this will be the time consuming part, having to move every existing string out into a compatible json file and using that as reference instead: https://docs.weblate.org/en/latest/formats/i18next.html#js-i18next

You can check out the weblate docs for an overview to see how you want to set things up. I'm not 100% sure, but I believe once I set up Weblate, I just import the base translation file (english, that you will make) and it will set all base strings up for translation process.

Thanks for looking into contributing! Let me know if you decide to start work on this we can discuss if you have any questions and I can assign you.

deadLocks21:
Hi @IRHM! Interested in contributing to this.

Quick question before I start: are we aiming for an instance-wide language setting (configured during the setup of the instance, same language for everyone), or a per-user language preference with dynamic switching on the frontend?

The i18next setup would differ quite a bit depending on the answer. Let me know!

luk6173:
I would like to see german language in Watcharr, what's the status?

Issue #772, opened 2025-01-20
Fetch texts in specified foreign language


Seyloria:
Hi there,

thx for your insane work so far.

Wouldn't it be a very usefull feature to be possible to fetch the overview and episode texts in your foreign language?

If there was a selection menu in the settings, similar to the streaming services country selection, all texts related to the movies and shows could be fetched and displayed in spanish/german/french for example.(i guess you are using tmdb as metadata provider in the background?!)

This would expand the potential audience for this tool by alot.

Translation of the UI would be nice aswell, but its not that important(could help translating into german if wished).

nash68:
I guess this is the parameter : 
	// Query params
	params := url.Values{}
	params.Add("api_key", getTMDBKey())
	**params.Add("language", "en-US")**
	for k, v := range p {
		params.Add(k, v)
	}

rkolja76:
Would love German support!

fjacques-2025:
I'd like to work on this (for French in my needs) — a configurable metadata language so overviews / titles
(and posters) are fetched from TMDB in the chosen language.

Approach I have in mind, in two steps:
1. A **global server option** (e.g. `TMDB_LANG`) so newly-fetched content comes in the
   selected language, with an English fallback when a translation is missing. Small and
   self-contained.
2. **Refreshing already-stored content** (existing overviews / titles / posters) into the
   new language — this overlaps the "refresh our own data" work (#1048 / #1049), so it may
   be better aligned with that than done as a one-off.

I'd start with (1). Does that fit with how you're thinking about #801 and the data-refresh?
Happy to open a draft PR.

PR #1079, opened 2026-07-25
feat: configurable metadata language (TMDB_LANG) with English fallback

fjacques-2025:
Implements #772 — fetch content metadata (titles, overviews, posters) from TMDB in a
chosen language.

## What
- New server setting **`TMDB_LANG`** (default `en-US`), edited from the server dashboard
  via a **language dropdown** (populated from TMDB primary translations). Selecting the
  **Default Country pre-fills** the matching language (still overridable).
- **English fallback**: TMDB returns empty fields when a translation is missing, so
  titles / overviews / posters are backfilled from `en-US`. Applied on **detail pages**,
  on **add/import**, and in **search results** (one extra en-US search, only when needed).
- New endpoint `GET /content/languages`.

## Relationship to #801
This is the **metadata** half (content coming from TMDB), independent of UI translation
(#801, Weblate/i18next).

## Design note / open question
The language is a **global server setting** here (simplest; my use case has a single
active user). #772 hints at a per-user selector (like the region dropdown): that's more
involved, because stored content (the `Content` table) is shared across users, so
per-user metadata would need per-language storage. I'm happy to take it that way, or to
align with how you're thinking about #801 / the data-refresh work — let me know your
preference.

## Notes
- Already-stored content keeps its previous language until re-fetched (opening a title
  refreshes it; a full re-import converts everything). A dedicated "refresh all" is
  intentionally left out of this PR.
- Opened as a **draft**: sharing a working solution (running on my own instance) for
  direction before any polishing.

## Testing
With `TMDB_LANG=fr-FR`: search, detail pages, and add/import all return French, with an
English fallback for untranslated entries.

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -66,6 +66,12 @@ type ServerConfig struct {
 	// If unprovided, the default Watcharr API key will be used.
 	TMDB_KEY string `json:",omitempty"`
 
+	// Optional: Language for TMDB metadata (titles, overviews, posters).
+	// ISO 639-1, optionally with a region (e.g. "fr-FR"). Defaults to "en-US".
+	// Note: changing this affects newly-fetched content; already-stored content
+	// keeps its previous language until refreshed.
+	TMDB_LANG string `json:",omitempty"`
+
 	// Optional: Point to Plex install to enable plex features.
 	PLEX_HOST string `json:",omitempty"`
 
@@ -106,6 +112,7 @@ func (c *ServerConfig) GetSafe() ServerConfig {
 		JELLYFIN_HOST:   c.JELLYFIN_HOST,
 		USE_EMBY:        c.USE_EMBY,
 		TMDB_KEY:        c.TMDB_KEY,
+		TMDB_LANG:       c.TMDB_LANG,
 		PLEX_HOST:       c.PLEX_HOST,
 		PLEX_MACHINE_ID: c.PLEX_MACHINE_ID,
 		DEBUG:           c.DEBUG,
@@ -135,6 +142,8 @@ func (c *ServerConfig) Get(s string) (ServerConfigGetByName, error) {
 		return ServerConfigGetByName{Value: c.SIGNUP_ENABLED}, nil
 	case "TMDB_KEY":
 		return ServerConfigGetByName{Value: c.TMDB_KEY}, nil
+	case "TMDB_LANG":
+		return ServerConfigGetByName{Value: c.TMDB_LANG}, nil
 	case "PLEX_HOST":
 		return ServerConfigGetByName{Value: c.PLEX_HOST}, nil
 	case "PLEX_MACHINE_ID":
@@ -161,6 +170,8 @@ func (c *ServerConfig) UpdateConfig(k string, v any) error {
 		c.SIGNUP_ENABLED = v.(bool)
 	} else if k == "TMDB_KEY" {
 		c.TMDB_KEY = v.(string)
+	} else if k == "TMDB_LANG" {
+		c.TMDB_LANG = v.(string)
 	} else if k == "DEBUG" {
 		c.DEBUG = v.(bool)
 		logging.SetLevel(c.DEBUG)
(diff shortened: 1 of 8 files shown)


fjacques-2025:
Rebased onto v4.2.x + migrated to the new `req` helper, running daily (French metadata with English fallback). Happy to push it for review, or align it with #801 — your call. 🙂

matimb0:
Hey! I think this PR is a really good idea — I actually came across it because I was planning to implement something similar.

Have you considered using a separate `contents_translation` table for this? My thought would be to keep `contents` always in English as the canonical source and store localized fields in something like:

`contents_translation(id, content_id, language, title, overview, poster_path, ...)`

I think this could be useful because currently the language isn't stored alongside the localized content, so later it may be difficult to determine which language the cached values are actually in. Also, because of the fallback mechanism, a single `contents` row could potentially contain values from two different languages. This could make the language of individual fields even harder to determine, especially if `TMDB_LANG` changes.

With a separate table, we could keep the English content untouched, support multiple translations, and fall back to English for missing fields. It could also make supporting different languages per user easier in the future.

The main reason I'm suggesting this is that I've also been thinking about working on a proper i18n system for Watcharr. I'd like to ask whether this is something the project would be interested in, and I'd be happy to work on it if so. For that, I think keeping the canonical content and its translations separately would be essential.

fjacques-2025:
Hey, sorry for the slow reply — I was away on holiday. Thanks for the thoughtful comment.

You're right on all three counts, and I'd go further: I just re-checked the code and the language isn't only absent from `contents`, it's also absent from the in-memory cache key (`CreateCacheKey("ShowDetails", id, country, params)` — the language is added at request time, not through `Params`). Harmless while `TMDB_LANG` is fixed at boot, but it's the same blind spot.

To be clear about what this PR is: deliberately the short path for a single-language instance. One global server option, fallback applied **per field** on `ShowDetails`/`MovieDetails`, cached wherever the existing content cache already writes. It's been running on my instance in daily use for a couple of months (French metadata, English fallback) and it does the job, but it doesn't answer "what language is this row in?", and it can't do per-user language. Your `contents_translation` design does both, and canonical English is the right anchor. No argument from me — I'd rather see that land than this.

A few things I ran into that might save you some time, whichever shape it takes:

- **Fall back field by field, not row by row.** TMDB very often has a translated title and an empty overview, or a translated overview and no localized poster. A whole-row fallback throws away good translations; no fallback shows blank overviews. Per-field is ugly, but it's what the data actually looks like.
- **Posters are part of this.** `poster_path` is localized by TMDB and goes missing more often than the text does, so it needs the same fallback treatment as the text fields.
- **Search is a separate surface** and this PR doesn't cover it (three generic `SearchX` functions). It's transient enough that I let it go, but a proper i18n system probably shouldn't.
- **Seasons/episodes are another one** — separate endpoint, separately cached (24h here).
- **Existing rows need a plan.** Instances that have already run with a non-English `TMDB_LANG` (mine included) have `contents` rows in that language with nothing recording it. A migration can't tell them apart from English ones, so it's either "assume English and let it be corrected on the next fetch", or re-fetch. Worth deciding early rather than late.

On the bigger question: I'd support a proper i18n system. One distinction worth making explicit first, because the two are easy to conflate: **UI string translation** (nothing exists for it today) and **metadata language** (what this PR does). They can be built independently, and it's the second one your table solves.

Beyond that it's an architecture decision, and it needs @IRHM's direction before either of us builds on it. @IRHM — could you give us a steer? Whether you want localized metadata at all, and if so whether it should go through a separate translations table as @matimb0 suggests, decides whether this PR is the right shape or should make way for that work. Either answer is useful; what neither of us can do is guess.

Either way, don't treat this PR as something to work around. Happy to close it in favour of your approach, or to keep it alive as a stopgap — I'm running it on my own instance regardless, so I'm not blocked. Just tell me which is more useful to you. 🙂

matimb0:
Thanks a lot for the detailed response and for all the helpful pointers you mentioned in your reply.

And you're absolutely right about the distinction between UI string translation and metadata language. When I was thinking about i18n, I had both aspects in mind as part of a more complete overall concept. One of the reasons I came up with the `contents_translation` table in the first place was while thinking about how a UI language switch should work. If we allow users to change the UI language, we'd also need a reliable way to switch the metadata language accordingly.

Regarding the PR, I think it makes sense to leave it open for now and wait for @IRHM thoughts on the overall direction. Once there's a decision on that, we can figure out together what makes the most sense for this PR.

PR #1080, opened 2026-07-25
feat: quick media-type toggle (All/TV/Movies) with per-mode filters & sort

fjacques-2025:
Adds a small, always-visible toggle at the top of the watched list —
**All / TV Shows / Movies** (reusing the search's `film` / `tv` icons) — so you can
switch between shows and movies in one click instead of opening the filter menu. It
drives the existing type filter, so it works with everything already there.

## What makes it useful day-to-day
- **Each mode remembers its own status filter and sort order.** For example: TV Shows →
  only *watching*, Movies → everything, each with its own sort. Switching modes restores
  that mode's view instead of clobbering it. (Before, applying a *watching* filter to see
  in-progress shows would also hide all your finished movies — the two filters are ANDed.)
- Everything is persisted to `localStorage` and rehydrated on load.

## Implementation
- `store.svelte.ts`: `filterModes` / `sortModes` (status + sort per mode), a
  `setWatchedListMode(mode)` helper, and `modeOf()` to derive the current mode from the
  type filter. The `activeFilters` / `activeSort` setters remember the current mode's
  values; rehydrate restores them.
- `(app)/+page.svelte`: the toggle UI + styling (mirrors the search `MediaTypeFilter`).

## Notes
- Per-mode memory starts empty, so the first time you set a filter/sort in a mode it's
  remembered from then on — no behaviour change until you opt in.
- Running on my own instance in daily use. Happy to adjust naming/placement.

### DIFF
--- a/src/routes/(app)/+page.svelte
+++ b/src/routes/(app)/+page.svelte
@@ -7,7 +7,7 @@
 	import Spinner from "@/lib/Spinner.svelte";
 	import infScroll from "@/lib/util/infScroll";
 	import paginatedLoader from "@/lib/util/paginatedLoader.svelte";
-	import { clearActiveFilters, store } from "@/store.svelte";
+	import { clearActiveFilters, setWatchedListMode, store } from "@/store.svelte";
 	import type { Media } from "@/types";
 	import axios, { type GenericAbortSignal } from "axios";
 	import { onDestroy, untrack } from "svelte";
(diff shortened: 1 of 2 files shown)


fjacques-2025:
Rebased onto the current `dev` (v4.2.x) and migrated to the new `req` fetch helper — running it daily with no issues. Happy to push the updated branch; is there interest in merging this? 🙂

LeonNamowitz:
You know I like this idea. I am straight up stealing it, except I'm using 2 buttons, one for "Watchlist" and the other for "Recently Watched" because I tend to switch between these 2 using the filters and sort all the time.

PR #1106, opened 2026-08-22
Add "allowfullscreen" attribute to YouTube trailer iframe

LeonNamowitz:
### Changes made

1 word change. Just added the `allowfullscreen` attribute to the trailer iframe on content pages.

---

Perhaps there is some reason why this shouldn't be a thing, or I'm literally the only person actually watching trailers through this, but it's been driving me insane that I can't fullscreen the trailers using the iframe..

### DIFF
--- a/src/lib/content/VideoEmbedModal.svelte
+++ b/src/lib/content/VideoEmbedModal.svelte
@@ -43,6 +43,7 @@
 				frameborder="0"
 				width="100%"
 				height="100%"
+				allow="fullscreen"
 				allowfullscreen
 			></iframe>
 		</div>
--- a/src/lib/content/VideoEmbedModal.svelte
+++ b/src/lib/content/VideoEmbedModal.svelte
@@ -43,6 +43,7 @@
 				frameborder="0"
 				width="100%"
 				height="100%"
+				allowfullscreen
 			></iframe>
 		</div>
 	{/if}


IRHM:
Thanks for the patch @LeonNamowitz 

I don't see any reason why it shouldn't be allowed! I'll have a look at merging this soon.

Note for future maybe: Turns out the `allowfullscreen` property is deprecated, but the replacement (`allow="fullscreen *"`) isn't supported in all browsers yet, so we can come back to that in the far future when/if this breaks: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe#allowfullscreen

LeonNamowitz:
I see. Maybe we should do both for compatibilities sake? Saw this used a couple times now as well.
I don't think we need to wildcard the domains here.

```html
<iframe
	title="Video Embed"
	src={embed}
	frameborder="0"
	width="100%"
	height="100%"
	allow="fullscreen"
	allowfullscreen
></iframe>
```

No hurries, thanks for getting back to quickly.

IRHM:
@LeonNamowitz I don't see the harm in adding it as a future proof, but I guess that future is probably 10 years away, in which is plenty of time to migrate when the new option is more widely adopted, so I'm okay leaving it out for now.

I shall leave the decision up to you!

PR #1109, opened 2026-08-27
Prevent zooming while searching on an iPhone

matimb0:
The iPhone zooms into an input field when the font size is less than 16px.
In my opinion, that doesn't look very good.

### Changes made

Set the font-size to 16px for the input field.

### DIFF
--- a/src/routes/(app)/+layout.svelte
+++ b/src/routes/(app)/+layout.svelte
@@ -552,6 +552,7 @@
 
 		input {
 			width: 100%;
+			font-size: 16px;
 			font-weight: bold;
 			text-align: center;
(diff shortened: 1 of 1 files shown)


IRHM:
@matzewr Hey, is this issue still effecting you or have you solved it in your browser?

matimb0:
@IRHM Yes, the problem still exists.
I just didn't think about the other input fields that also zoom in (login, comments, settings, etc.).
So we'd have to set the font-size of every input and textarea to 16px globally. But that will probably throw off the layout.
That’s why I think the easiest way to fix this would be to use `<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">`.
What do you think?

PR #1114, opened 2026-09-01
feat(watched): enable watched date in /watched/* API

matimb0:
I'm currently building a Watcharr Scrobbler browser add-on based on the Universal Trakt Scrobbler (https://github.com/trakt-tools/universal-trakt-scrobbler).
This needs to be able to mark individual episodes as watched and pass a watched date when doing so.

### Changes made
Added support for external watched date syncing.

### DIFF
--- a/server/domain/watched.go
+++ b/server/domain/watched.go
@@ -202,7 +202,8 @@ type WatchedUpdateRequest struct {
 	Pinned         *bool                `json:"pinned" `
 	// Allow the added activity count as play?
 	// If the activity was going to count, this can stop it.
-	LetCountAsPlay *bool `json:"letCountAsPlay"`
+	LetCountAsPlay *bool     `json:"letCountAsPlay"`
+	WatchedDate    time.Time `json:"watchedDate,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


matimb0:
It ended up being more than I first thought ;)
Having worked my way through the entire API, I would like to make the following changes.
With these changes, a `watched date` can now be passed consistently throughout the entire /watched API.
I have thought long and hard about whether – when an watched date is passed in – I should display the resulting activity using the type import, but personally I preferred the solution involving the added type; this way, the activity remains clearer.

As I said, the change is based on the fact that I wrote a Watcharr scrobbler browser add-on that lets you import your Netflix and Prime history into Watcharr.
Unfortunately, this doesn't work via /import; that method would already allow you to pass in the watched date.

As far as I can tell, this shouldn't affect any current projects. Otherwise, I would have consulted you beforehand.

@IRHM please let me know if I’ve missed anything or if you disagree with this entirely.
And many thanks for your work. I think Watcharr is a truly wonderful project, which, in my opinion, is the best open-source alternative to Trakt.

